PR #8, opened 2020-03-27
dev in conda

### LINKED ISSUE #7: development in conda for aghast (ROOT)
@henryiii, we might be forced to use conda env if we want to use ROOT in aghast, as mentioned in [#28](https://github.com/scikit-hep/aghast/issues/28). Thus, we need a conda development guideline (and maybe even remove the venv guideline).

LovelyBuggies:


LovelyBuggies:
CI / pre-commit failing...

LovelyBuggies:
@henryiii Could you help me with this CI problem? I don't know why I got this.

.github/CONTRIBUTING.md
```diff
@@ -1,12 +1,24 @@
...
+You can set up a development environment using PyPI.
 
 ```bash
-python3 -m venv .env
-. .env/bin/activate
-pip install -e .[dev]
+$ python -m venv .env
+$ source .env/bin/activate
+(.env)$ pip install -r requirements.txt
+(.env)$ pip install -e .
```
henryiii: This should remain `-e .[dev]`, and we should not add the requirements.txt. The requirements.txt is wrong, too, it should not install boost-histogram version right before you install a boost-histogram from source.

.github/CONTRIBUTING.md
```diff
@@ -1,12 +1,24 @@
...
+(.env)$ python -m ipykernel install --user --name hist
 ```
 
-You should have pip 10 or later.
```
henryiii: You still need pip 10 or later.

.github/CONTRIBUTING.md
henryiii: Packages should not have a requirements.txt, see the discussion on scikit-hep/particle's recent issue.

LovelyBuggies: > we should not add the requirements.txt

Yes, packages should not have requirements.txt. But I think requirements.txt might be ok for development. See the [aghast #32](https://github.com/scikit-hep/aghast/pull/31), and these changes are made for CONTRIBUTING.md.

> The requirements.txt is wrong, too, it should not install boost-histogram version right before you install a boost-histogram from source.

If so, is it ok to remove boost-histogram requirement?

henryiii: Please make this dev-requirements.txt at least; I still rather not have this though since it can get out of sync with setup.py, while the `[dev]` cannot.

It is okay to assume developers know a bit more than normal users.

LovelyBuggies: All right. I will do what I did for aghast (adding <-test> suffix for requirements.txt and environment.yml).

LovelyBuggies:
@henryiii Anything else to change?

LovelyBuggies:
Btw, should this -- conda dev support -- be added for boost-histogram?

dev-environment.yml
```diff
@@ -0,0 +1,9 @@
+name: hist
+channels:
+  - conda-forge
+dependencies:
+  - pip >=18
+  - root
+  - pip:
+    - -r dev-requirements.txt
+    - -e .[dev]
```
henryiii:
```suggestion
    - -e .
```

If there are any dev requirements, these need to go in as conda requirements if at all possible, not as pip ones.

dev-environment.yml
```diff
@@ -0,0 +1,9 @@
+name: hist
+channels:
+  - conda-forge
+dependencies:
+  - pip >=18
+  - root
```
henryiii:
```suggestion
  - root >=6.20.0
```

Might speed up the solver a bit.

henryiii:
> Btw, should this -- conda dev support -- be added for boost-histogram?

No, let's keep boost-histogram simple. And it's really only a special case for the 2-3 people who actually download the repository - for the vast majority of users, they will be running `pip install boost-histogram` or maybe `pip install boost-histogram[dev]`, and will not have access to these files.

environment.yml is special, since mybinder.org picks it up; so that is always the requirements to run the example notebooks.

dev-environment.yml
henryiii: Note that this is really here just to support the tests with ROOT.

LovelyBuggies:
@henryiii Agreed all your commits above.

dev-environment.yml
```diff
@@ -0,0 +1,9 @@
+name: hist
+channels:
+  - conda-forge
+dependencies:
+  - pip >=18
+  - root >=6.20.0
+  - pip:
+    - -r dev-requirements.txt
```
henryiii:
```suggestion
```

This environment should have conda dependencies, not PyPI ones. dev-requirements is PyPI.

henryiii: The only exception is if requirements are *only* available through pip, but I don't think we have any of those.

LovelyBuggies: Mean that we need to `pip install -r dev-requirements.txt` in the conda environment manually?

LovelyBuggies: > Mean that we need to `pip install -r dev-requirements.txt` in the conda environment manually?

If so, I add a this line to CONTRIBUTING.md instructions as you can see in https://github.com/scikit-hep/hist/pull/8/commits/a57d8280743c59bc0ae0d544c108c73992cd6073

henryiii: No. PyPI and Conda are different. You have to list the packages again here in this file if you want them - since they aren't even guaranteed to have the same Conda name as PyPI name (though often they do).

LovelyBuggies: > they aren't even guaranteed to have the same Conda name as PyPI name (though often they do)

👍 You are right! And some even do not have conda installation like mplhep.

LovelyBuggies: No dev-requirements.txt anymore.

LovelyBuggies:
LGTM, able to merge?

.github/CONTRIBUTING.md
```diff
@@ -1,12 +1,26 @@
...
+
+You can set up a development environment using PyPI.
 
 ```bash
-python3 -m venv .env
-. .env/bin/activate
-pip install -e .[dev]
+$ python3 -m venv .env
+$ source .env/bin/activate
+(.env)$ pip install -r dev-requirements.txt
```
henryiii, line 11:
```suggestion
```

henryiii:
This should be merged, not sure why it's "open".

PR #13, opened 2020-05-25
NamedHist Init

LovelyBuggies:


LovelyBuggies:
@henryiii This PR is to construct NamedHist obj and func for Hist.

Plan:

- Basic functionality of NamedHist: tested on the notebook
- Typing hints in it: tested with mypy pre-commit
- Unit Test: tested with pytest

src/hist/named.py
```diff
@@ -1,5 +1,34 @@
 from .core import BaseHist
+import numpy as np
 
 
 class NamedHist(BaseHist):
-    pass
+
+    # ToDo: judge whether have no-named axes when initialization
```
LovelyBuggies, line 7: Basic functionality in progress

tests/test_basic.py
```diff
@@ -3,3 +3,12 @@
...
 def test_version():
     assert hist.__version__ is not None
+
+
+def test_basic_usage():
+    """
+        Test whether the function of boost-histogram can be used normally.
+    """
+
+    # ToDo:
```
LovelyBuggies, line 13: @henryiii how to test here?

tests/test_named.py
```diff
@@ -1,16 +1,57 @@
...
+        elif idx == 4:
+            assert h[idx] == h[{0: idx}] == h[{"x": idx}] == 1
+        else:
+            assert h[idx] == h[{0: idx}] == h[{"x": idx}] == 0
 
-    assert h[2] == 0
-    assert h[3] == 2
-    assert h[4] == 1
-    assert h[5] == 0
+    h = NamedHist(
```
LovelyBuggies, line 21: @henryiii Should include other types of axes?

tests/test_basic.py
LovelyBuggies: Though this pr is not responsible for this test: this is a PR for NamedHist.

LovelyBuggies:
I remove 3.6 CI tests for the reason that annotation package cannot be found.

tests/test_basic.py
LovelyBuggies: Alright, let's test those basics in the future PRs :)

src/hist/_internal/axis.py
```diff
@@ -1,22 +1,34 @@
+from __future__ import annotations
```
henryiii, line 1:
```suggestion
```

Let's drop this one line and be 3.6+ compatible. There are a lot of 3.6 users out there. With end of life coming in September for 3.5, and NEP 29 already at 3.6+, we probably shouldn't bother with 3.5 support.

.github/workflows/ci.yml
```diff
@@ -24,7 +24,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-        - 3.6
```
henryiii, line 27: This should return with the one change below.

src/hist/named.py
```diff
@@ -1,5 +1,80 @@
 from .core import BaseHist
+import numpy as np
```
henryiii, line 4: We may move some/most of the logic to BaseHist later, with this just wrapping it with errors if a user does not use names, but this is great for now!

src/hist/_internal/axis.py
henryiii: It will make typing a little uglier down the line when we do fancier things, but mostly will work.

src/hist/named.py
LovelyBuggies: okay.

LovelyBuggies:
@henryiii Some new commits, I think the tests for NamedHist are detailed enough. LGTM, PTAL. Take a review before merging it :)

PR #16, opened 2020-06-04
Hist init

LovelyBuggies:


LovelyBuggies:
@henryiii Ready for review.

setup.cfg
```diff
@@ -46,14 +46,16 @@ package_dir =
...
   boost-histogram ==0.7.0
+  matplotlib >=2.0
+  scipy >=1.4.1
 
 [options.extras_require]
 dev =
   pytest >=4.6
 test =
   pytest >=4.6
 mpl =
```
henryiii, line 57: You don't actually need a mpl extra if mpl is already required to install.

LovelyBuggies: Not sure. IIRC, CI failed due to this line `MPL >=2.0`.

PR #18, opened 2020-06-10
mv hist funcs to BaseHist

LovelyBuggies:


LovelyBuggies:
This PR moves all funcs of Hist to BaseHist and add some tests (whether BaseHist works, whether pull plot tests can pass in NamedHist). Note that I enable the access of BaseHist for test_basic.py.

PR #21, opened 2020-06-17
nino/docs

LovelyBuggies:
`make html` under /docs folder to build docs. Considering the usage of hist is in flux, I cannot guarantee the correctness of original code snippet and Jupyter notebooks, so I have not provided `User Guide` and `Examples` up to now (though I can do this, there will be many fine-tunes). I plan to set aside some time for these docs at the end of GSoC. Then the doc could be both comprehensive and coherent. And it also helps me to recall back when writing my GSoC final article and maybe *presentation*.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+SOURCEDIR     = source
+BUILDDIR      = build
+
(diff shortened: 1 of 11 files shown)


LovelyBuggies:
@henryiii Maybe you can tell me which to include in Development (development guidelines) section or add some content by yourself :)

henryiii:
> But I think it's reasonable to wait until some functionalities are stable at least.

Yes, fully agree!

henryiii:
> Maybe you can tell me which to include in Development

Probably something similar to https://scikit-hep.org/developer/intro ? Doesn't have to be added right away. Also, do you think we could add to the Scikit-HEP dev pages with some info on setting up Sphinx?

PR #22, opened 2020-06-17
nino/Hist-patch4

LovelyBuggies:
## Mini-Project

- [X] implement projection
- [x] test projection
- [x] implement plot
- [x] test plot
- [x] add_axes issue (*ToDo comments*) 

## Legacy

- [ ] minimal instead of curve_fit
- [ ] disable other types of axes in pull_plot and plot
- [ ] disable projection via index for NamedHist (*ToDo comments*)
- [ ] add return types for `__g/setitem__`
- [ ] `project` returns a Union of the histogram type and the contents type.


henryiii:
For projection, just forward to the build in `.project` method, just doing the name conversion, like you do in the other methods.

src/hist/core.py
```diff
@@ -11,19 +11,26 @@
...
         for ax in self.axes:
             if ax.name in self.names:
                 raise Exception(
-                    "Hist instance cannot contain axes with duplicated names."
+                    "BaseHist instance cannot contain axes with duplicated names."
                 )
             else:
                 self.names[ax.name] = True
 
+    def project(self, *args: Tuple[int]):
```
LovelyBuggies, line 27: How to deal with the pre-commit failure.

src/hist/named.py
```diff
@@ -12,6 +13,25 @@ def __init__(self, *args, **kwargs):
...
+            for index, axis in enumerate(self.axes):
+                if name == axis.name:
+                    indices += (index,)
+                    find = True
+                    break
+
+            if not find:
+                raise ValueError("The axis names could not be found.")
+
+        return super(Histogram, self).project(*indices)
```
LovelyBuggies, line 33: **How to skip the `BaseHist` and directly use `bh.Histogram`**?

LovelyBuggies: When using `super().project(*indices)`, pytest fails.

src/hist/core.py
henryiii: `Tuple[Union[str, int]]` is the correct signature here. We may have to disable this one check on one line for NamedHist, as it will have more restrictive type signatures than Hist, which it's based on. There's a way to do that which I can look up.

src/hist/named.py
henryiii: Super goes one up. If you want to go up further, just call it directly. So `Histogram.project(self, *indices)`. However, you don't actually want that; NamedHist should: check to see if all arguments are strings - if one or more are ints, it should fail. Then it should just pass on to Hist: `super().project(*args)`. Hist should take strings or ints.

src/hist/named.py
```diff
@@ -21,18 +41,21 @@ def fill(self, *args, **kwargs):
...
         for name, val in kwargs.items():
+            find = False
             for index, axis in enumerate(self.axes):
                 if name == axis.name:
                     indices.append(index)
                     values.append(val)
+                    find = True
                     break
 
+            if not find:
```
henryiii, line 52: You can actually just make this an else class *on the for loop*, and drop the find variable. The else clause of a for loop runs when the loop was never broken.

henryiii: https://book.pythontips.com/en/latest/for_-_else.html#else-clause

src/hist/core.py
```diff
@@ -11,19 +11,26 @@
...
+            Initialize BaseHist object. Axis params must contain the names.
         """
 
         super().__init__(*args, **kwargs)
         self.names: dict = dict()
         for ax in self.axes:
             if ax.name in self.names:
                 raise Exception(
-                    "Hist instance cannot contain axes with duplicated names."
+                    "BaseHist instance cannot contain axes with duplicated names."
```
henryiii, line 22:
```suggestion
                    f"{self.__class__.__name__} instance cannot contain axes with duplicated names."
```

henryiii: You can use `type(self).__name__` if you like. In Python 3, `type(self) == self.__class__` always, and it is mostly always true in Python 2 (as long as you use new-style classes, which everyone should). I use the syntax above because it works in the old style `"{self.__class__.__name__}".format(self=self)` syntax, whereas calling a function like type does not. f-strings don't care. They are also faster.

src/hist/core.py
```diff
@@ -11,19 +11,26 @@
 class BaseHist(Histogram):
     def __init__(self, *args, **kwargs):
         """
-            Initialize Hist object. Axis params must contain the names.
+            Initialize BaseHist object. Axis params must contain the names.
```
henryiii, line 14:
```suggestion
            Initialize BaseHist object. Axis params can contain the names.
```

Only `NamedHist` *requires* names. They are optional in BaseHist.

src/hist/core.py
LovelyBuggies: Okay. But it's incompatible with `NamedHist.fill()`, where users are only allowed to fill by names.

src/hist/named.py
LovelyBuggies: Yeah, we should allow NamedHist to use indices to project, which is different from the fill method. This is because we need to use super(), but NamedHist must be derived from BaseHist. Seems like Python's derivation must be an addition, and NamedHist has to accept the feature of its father class if we want to succeed the features of its grandfather.

LovelyBuggies: Ah, at last. Find the problem. it's risky to do that.

> NamedHist should check to see if all arguments are strings - if one or more are ints, it should fail.

**We must allow projection by the index of NamedHist**, if we don't plan to change anything of underlying BH or get_item. That's because get_item will use this property. More generally speaking, Python's derivation is an addition, if we are forced to make some minus, then some manners might be broken. 

So the current idea is, BasedHist allows projection by the index, and NamedHist **add** the feature that allows projection by the names.

src/hist/core.py
LovelyBuggies: A typo 😊.

src/hist/named.py
henryiii: NamedHist *only* forces names.  Normal Hist prefers names, but works with both.

henryiii: > That's because get_item will use this property.

No, it calls project on the c++ class (unless it doesn't, in which case it's a bug), so it's okay if the python class doesn't support integers. But that is why you have to force this on MyPy; it is unexpected if a child class doesn't accept something a parent does. But in this case, I think it's okay.

henryiii: https://github.com/scikit-hep/boost-histogram/blob/a12244f5daf81d5e619cefdd70d0d2ad859ca39d/src/boost_histogram/_internal/hist.py#L574

This line returns the c++ class, and not a python wrapper.

henryiii: Static types would have made this easier to read in boost histogram!

src/hist/core.py
```diff
@@ -5,25 +5,38 @@
...
             else:
                 self.names[ax.name] = True
 
+    def project(self, *args: Tuple[Union[int, str]]):
+        """
+        Projection of axis idx.
+        """
+
+        if len(args) != 0:
+            if not all(isinstance(x, int) for x in args):
```
henryiii, line 33: This needs to convert any names to axis integers

src/hist/named.py
LovelyBuggies: I made a new commit https://github.com/scikit-hep/hist/pull/22/commits/4661c9a840bc0df84807010672850f1e05364990 then and you can see that we cannot make subtraction for `NamedHist.project()`.

LovelyBuggies: This road seems to be blocked.

src/hist/named.py
```diff
@@ -10,7 +11,28 @@ def __init__(self, *args, **kwargs):
...
+            raise Exception(
+                f"Each axes in the {self.__class__.__name__} instance should have a name"
+            )
+
+    def project(self, *args: Tuple[Union[int, str]]):
+        """
+        Projection of axis idx.
+        """
+
+        if all(isinstance(x, int) for x in args):
```
henryiii: Move this function to BaseHist and then the function here will just have:

```python

if all(isinstance(x, str) for x in args):
    return super().project(*args)
else:
    raise ...
```

LovelyBuggies: That's exactly what I do in my new commit.

src/hist/named.py
henryiii: Bug in boost-histogram then. Let's disable the check for no ints for one (in NamedHist) and tomorrow I'll fix the problem in boost-histogram. Add a comment with a TODO in the Hist src.

LovelyBuggies: Okay!

LovelyBuggies: Don't you think disable projection by names for BaseHist is more reasonable? As methods related to names seem to be exclusive for NamedHist.

henryiii: Looks like I wrapped it here: https://github.com/scikit-hep/boost-histogram/blob/a12244f5daf81d5e619cefdd70d0d2ad859ca39d/src/boost_histogram/_internal/hist.py#L304. Need to think about how to do this consistently.

LovelyBuggies: Afterall, go ahead, these are all based on correct BH :)

henryiii: No! Names are a feature for all of Hist. We don't want to discourage name use. The only reason we have NamedHist is we have one user who wants to enforce names always. Most users will mix the two. Add names, but also use position when it's more convenient, like in indexing. 

Maybe we should rename NamedHist to ForcedNamedHist. :)

LovelyBuggies: Okay, I see.

LovelyBuggies:
@henryiii Could you plz take a look at the last item in the mini-project -- `fig.add_axes()` issue. I reconstruct the tests and `plot` is okay as U can see in the notebooks.

henryiii:
```python
h.plot() # -> Calls plot1d or plot2d based on h.ndim, removes any extra stuff
h.plot1d() # -> The 1D part of h.plot
h.plot2d() # -> Adds the 2 1D plots by default, but can be removed
h.pull_plot() # requires function arguments
```

The plot impl should be easy:
```python
    def plot(self, *args, **kwargs):
        if self.ndim == 1:
            return self.plot1d(*args, **kwargs)
        elif self.ndim == 2:
            return self.plot2d(*args, topax=False, sideax=False, **kwargs)
        else:
            raise NotImplemented("Please project to 1D or 2D before calling plot")
```

src/hist/core.py
```diff
@@ -1,29 +1,170 @@
...
+                    raise ValueError("The axis names could not be found")
+
+            return super().project(*indices)
+
+        else:
+            raise TypeError(
+                f"Only projections by indices and names are supported for {self.__class__.__name__}"
+            )
+
+    def plot(
```
henryiii, line 52:
```suggestion
    def plot2d_full(
```

LovelyBuggies:
> * [x]  `h.plot() # -> Calls plot1d or plot2d based on h.ndim, removes any extra stuff`
> * [x]  `h.plot1d() # -> The 1D part of h.plot`
> * [x]  `h.plot2d() # -> Simple 2D plot`
> * [x]  `h.plot2d_full() # -> Adds the 2 1D plots`
> * [x]  `h.plot_pulls() # requires function arguments`
> 
> The plot impl should be easy:
> 
> ```python
>     def plot(self, *args, **kwargs):
>         if self.ndim == 1:
>             return self.plot1d(*args, **kwargs)
>         elif self.ndim == 2:
>             return self.plot2d(*args, **kwargs) # not plot2d_full
>         else:
>             raise NotImplemented("Please project to 1D or 2D before calling plot")
> ```

@henryiii Plot family complete. Few problems: 

- The implementation is a little bit verbose. To pass the test, I reuse lots of code, `plot` would be better to base on `plot1d` and `plot2d` (similar `plot_pull` would be based on `plot2d` at least).
- I am not sure about how to define the return type hints of `plot`.
- `ndim` cannot be found for some reason, so I used the `len(self.axes)`.

LovelyBuggies:
Plot family is almost completed. Few problems:

- [ ] The implementation is a little bit verbose. To pass the test, I reuse lots of code, plot would be better to base on plot1d and plot2d (similar plot_pull would be based on plot2d at least).
- [ ] I am not sure about how to define the return type hints of plot.
- [ ] ndim cannot be found for some reason, so I used the len(self.axes).

henryiii:
> ndim cannot be found for some reason

Ouch, that's my fault, I used the NumPy `.ndim` instead of the Boost.Histogram `.rank`. I've wanted to allow ndim on histograms, actually. Just opened an issue to see what Hans thinks.

henryiii:
> The implementation is a little bit verbose

I think we could merge as-is, then refactor later. Ideally, `plot_full` and `plot_pull` would use the simpler plots and then just extend them.

src/hist/core.py
```diff
@@ -2,29 +2,270 @@
...
             if ax.name in self.names:
                 raise Exception(
-                    "Hist instance cannot contain axes with duplicated names."
+                    f"{self.__class__.__name__} instance cannot contain axes with duplicated names."
                 )
             else:
                 self.names[ax.name] = True
 
-    def pull_plot(
+    def project(self, *args: Union[int, str]):
```
henryiii, line 28: This returns a Union of the histogram type and the contents type. I can show you how to set that up later.

src/hist/core.py
```diff
@@ -2,29 +2,270 @@
...
+
+    def plot2d_full(
+        self,
+        fig: Optional[matplotlib.figure.Figure] = None,
+        main_ax: Optional[matplotlib.axes._subplots.SubplotBase] = None,
+        top_ax: Optional[matplotlib.axes._subplots.SubplotBase] = None,
+        side_ax: Optional[matplotlib.axes._subplots.SubplotBase] = None,
+        **kwargs,
+    ) -> Tuple[
+        Optional[matplotlib.figure.Figure],
```
henryiii, line 160: Are these actually Optional? I would assume it would just return the passed in fig if fig is not None, or the new fig if it's not. So I don't think the returns here are optional.

src/hist/core.py
```diff
@@ -2,29 +2,270 @@
...
+            return super().project(*indices)
+
+        else:
+            raise TypeError(
+                f"Only projections by indices and names are supported for {self.__class__.__name__}"
+            )
+
+    def plot(
+        self, *args, **kwargs,
+    ):  # ToDo: how to add return type hints
```
henryiii, line 54: It's just a Union of the other two. I'd make a type alias, such as:

```python
Plot1D_RetType = Tuple[matplotlib.figure.Figure, matplotlib.axes._subplots.SubplotBase]
```

Then use it here in a `Union[Plot1D_RetType, Plot2D_RetType]`, as well as for the 1D return (also 2D).

src/hist/named.py
```diff
@@ -10,7 +11,22 @@ def __init__(self, *args, **kwargs):
...
+            raise Exception(
+                f"Each axes in the {self.__class__.__name__} instance should have a name"
+            )
+
+    def project(self, *args: Union[int, str]):
+        """
+        Projection of axis idx.
+        """
+
+        # ToDo: should work
```
henryiii, line 23:
```suggestion
        # ToDo: should work in boost-histogram 0.8.0
```

src/hist/core.py
henryiii: I think MyPy is smart enough that if you do:

```python
x: Optional[A]

... # Stuff

if x is None:
    x = A()
```

It will know from now on x is not Optional. If not, there are two options. One is use a different variable name for the non-optional value, and second is to place this after the location where it is no longer optional:

```python
assert x is not None
```

MyPy will then know for sure that x is no longer optional (Same thing works for unions and types - since Optional really is just a Union with None).

LovelyBuggies: Yean, it should not be optional :)

src/hist/core.py
LovelyBuggies: Okay, it works fine. But I wonder why I cannot use like this:

```python
Fig_Type = matplotlib.figure.Figure
Ax_Type = matplotlib.axes._subplots.SubplotBase
Plot2D_RetType = Tuple[Fig_Type, Ax_Type]
```

Looks like `Fig_Type` is not a **type** :)

src/hist/core.py
LovelyBuggies: okay.

Issue #5, opened 2020-03-17
Analysis shortcuts


henryiii:
This is to enable simple, quick analysis in a notebook or REPL. Here are current ideas (to be expanded):

#### Allow complex numbers for `bh.loc`

This would allow `bh.loc(1.5) + 2` to be written as `1.5j + 2`. This is easier to read and type.

#### Allow `sum` for `bh.sum`

We can check for the python `sum`, allowing `h[::bh.sum]` to be written as `h[::sum]`.

#### Allow auto full-range slicing

If someone does `h[bh.rebin(2)]`, that possibly could implement `h[::bh.rebin(2)]` automatically. This is a bit tricky (since single callables already are supported like `bh.loc`), but might be possible.

henryiii:
Note: these could be opt-in at first, since other ideas have been discussed. If they become popular and work well, they might even be able to be upstreamed into boost-histogram. :)

LovelyBuggies:
Seems like `h[::bh.sum]` is not support, currently.

LovelyBuggies:
> This would allow bh.loc(1.5) + 2 to be written as 1.5j + 2. This is easier to read and type.

@henryiii , what does `j` mean? Could it possibly be any other characters like `i`? And could the index overflow after adding 2 (`bh.loc(1.5) + 2` should be an index OR this is to add 2 to every location with 1.5)?

LovelyBuggies:
**Should the Hist object own a name? Should the transform be available for Hist object (for every axis)? ...**

pros:
- Users can directly change (perform) many axes at one time.
- ...

cons (challenges):
- Ambiguous name
- Fill method
- ...

henryiii:
> what does j mean? Could it possibly be any other characters like i?

It's using the fact that Python has built-in support for complex numbers; `2j+1` becomes `complex(2, 1)`. So we are stuck with j. Complex numbers can never be a valid 1D index (since they are 2D), so it's safe (IMO) to use them. The same trick is used by numpy in np.mgrid and several other places.

> And could the index overflow after adding 2 (bh.loc(1.5) + 2 should be an index OR this is to add 2 to every location with 1.5)?

This does what `bh.loc(1.5) + 2` already does; it looks up the bin number that contains 1.5, then it adds two to that number. If that overflows, then it overflows.

henryiii:
> Should the Hist object own a name 

[...]

PR #23, opened 2020-06-29
Nino/Complex-Index

LovelyBuggies:
Work for part of #5 .


henryiii:
`loc(v.real, v.imag)` should just work: https://github.com/scikit-hep/boost-histogram/blob/a12244f5daf81d5e619cefdd70d0d2ad859ca39d/src/boost_histogram/tag.py#L70

src/hist/core.py
```diff
@@ -241,3 +241,24 @@ def pull_plot(
...
 
         return fig, ax, pull_ax
+
+    def __getitem__(self, index):
+        """
+            Get histogram item.
+        """
+
+        if not hasattr(index, "__iter__"):
+            index = () + (index,)
```
henryiii, line 251:
```suggestion
            index = (index,)
```

src/hist/core.py
```diff
@@ -241,3 +241,24 @@ def pull_plot(
...
+
+        t: tuple = ()
+        for i, idx in enumerate(index):
+            if isinstance(idx, complex):
```
henryiii, line 255: Then just do:
```python
    t += (bh.loc(idx.imag, idx.real),)
```

Or, you could do this in one go:

```python
t = tuple((bh.loc(idx.imag, idx.real) if isinstance(idx, complex) else idx) for idx in index)
```

henryiii: You'll also want to add a check to see if it's a string, and then wrap in `loc` in that case too.

LovelyBuggies:
Able to merge. Btw, do I need to add tests for setitem?

src/hist/core.py
```diff
@@ -473,3 +473,47 @@ def plot_pull(
...
 
         return fig, ax, pull_ax
+
+    def __getitem__(self, index):
+        """
+            Get histogram item.
+        """
+
+        if isinstance(index, dict):
+            return super().__getitem__(index)
```
henryiii, line 483: Does the shortcut not work on dict items?

src/hist/core.py
```diff
@@ -473,3 +473,47 @@ def plot_pull(
...
+
+        if not hasattr(index, "__iter__"):
+            index = (index,)
+
+        t: tuple = ()
+        for idx in index:
+            if isinstance(idx, complex):
+                t += (loc(idx.imag, int(idx.real)),)
```
henryiii, line 491: Great, though I think maybe we should throw an error if the real part is not exactly an integer? Not sure, since it internally obviously it is an float.

src/hist/core.py
LovelyBuggies: Yes, the following parts, saying the `t: tuple`, will only use the keys of the `index`, if `isinstance(index, dict)` and pass them to `super.__getitem()`.

src/hist/core.py
LovelyBuggies: Yes, I am not sure how to deal with it, either. As the `2j+2` is passed as `2.0` sometimes.

LovelyBuggies:
> You might want to break up the tests into smaller functions.

Sure, I want! And it's a good manner. But I am not sure how to split the test_basic to small parts, i.e. where to split and what's 
 the split grained. Maybe this is not what this PR is focusing on and I am patient enough to wait for the pytests, so I do not plan to change it in this PR. I'll try to reconstruct test files again in the future after the shortcuts work (after the proxy work) is completed :)

src/hist/core.py
henryiii: It should be a float all the time. But you can check if `idx.real % 1 == 0` and raise an error if not.

henryiii:
I'd still like two changes: Pulling the ` if isinstance`'s out into a helper function, and applying them to dictionary values as well.

```python
def _loc_shortcut(x):
    if isinstance(x, slice):
        return slice(_loc_shortcut(x.start), _loc_shortcut(s.stop), _step_shortcut(x.step))
    elif isinstance(x, complex):
        return loc(...
    elif ...

def _step_shortcut(x):
    return x # for now, can check for complex and convert to rebin, maybe?

# Inside getitem/setitem:
   if isinstance(index, dict):
        return {key, _loc_shortcut(value) for key, value in index.items()}
   ...
   return tuple(_loc_shortcut(value) for value in index)
```

LovelyBuggies:
> I'd still like two changes: Pulling the ` if isinstance`'s out into a helper function, and applying them to dictionary values as well.
> 
> ```python
> def _loc_shortcut(x):
>     if isinstance(x, slice):
>         return slice(_loc_shortcut(x.start), _loc_shortcut(s.stop), _step_shortcut(x.step))
>     elif isinstance(x, complex):
>         return loc(...
>     elif ...
> 
> def _step_shortcut(x):
>     return x # for now, can check for complex and convert to rebin, maybe?
> 
> # Inside getitem/setitem:
>    if isinstance(index, dict):
>         return super().__getitem__({key, _loc_shortcut(value) for key, value in index.items()})
>    ...
>    return super().__getitem__(tuple(_loc_shortcut(value) for value in index))
> ```

@henryiii Follow your advice, I made some changes and get/setitem passed the tests basically. But there's a question, I did not figure out why we should have a func named `_step_shortcut()` for complex and rebin. We can simply use `x.step` for slice and `_loc_shortcut()` for complex numbers, can't we? What's the meaning of `_step_shortcut()`? Can you show me a situation that we should use it?

henryiii:
Currently, if you want to rebin, you have to do:

```python
h[::bh.rebin(2)]
```

One possible shortcut would be to allow complex steps to count as rebin:

```python
h[::2j]
```

In this case, `_step_shortcut` would convert pure imaginary complex numbers to `bh.rebin(int(v.imag))`.

henryiii:
I think this looks good now!

PR #24, opened 2020-06-30
Nino/hist patch5

LovelyBuggies:


LovelyBuggies:
just change len(self.axes) to self.rank

PR #27, opened 2020-07-06
Nino/index complex

LovelyBuggies:


src/hist/core.py
```diff
@@ -493,6 +495,19 @@ def _loc_shortcut(self, x):
...
             return x
 
+    def _step_shortcut(self, x):
+        """
+            Convert some specific indices to step.
+        """
+
+        if isinstance(x, complex):
+            if x.real != 0:
+                raise ValueError(f"The step should not have real part")
```
henryiii, line 505: Could you add the same check for fractional complex numbers as in loc, too?

LovelyBuggies: ok

Issue #29, opened 2020-07-07
Update to boost-histogram 0.8.0


henryiii:
We should update to boost-histogram 0.8.0. Things to watch for (that I currently recall):

* New Boolean axis type can now be used
* We can drop the workaround we had for `.project` (IIRC?)
* `bh.sum is sum` now
* `rank` should be changed to `ndim`, like NumPy (do a `git grep rank` from command line, or however you like to do it)

LovelyBuggies:
> New Binary axis type can now be used

Should we drop the `Bool` axis in hist?

> We can drop the workaround we had for .project (IIRC?)

I don't think we can drop it, since hist's `project` allow projection via names.

> bh.sum is sum now

This won't cause changes in hist package, currently.

> rank -> ndim

I'll make a change correspondingly.

henryiii:
> Should we drop the Bool axis in hist?

It should become a wrapper for Boolean and be called Boolean.

> I don't think we can drop it, since `hist`'s project allow projection via names.

I think the problem was we couldn't disable axes numbers in NamedHist without the fix?

LovelyBuggies:
> It should become a wrapper for Boolean and be called Boolean.

Okay. I'll change them.

> I think the problem was we couldn't disable axes numbers in NamedHist without the fix?

First, we could disable axes numbers in NamedHist just by **rewriting** the `project` function (eliminate the axes number projection in `BaseHist` and the other remain the same). Second, I am not sure what is the corresponding PR in BH that fix it (only BH PR 185 in the new release has something to do with `project`, but it seems not to be related to this issue).

henryiii:
PR 377.

LovelyBuggies:
There might be a bug:

```python
import boost_histogram # -V 0.8.0

h = bh.Histogram(bh.axis.Boolean(), bh.axis.Boolean(), bh.axis.Boolean()).fill(
            [True, True, True, True, True, False, True],
            [False, True, True, False, False, True, False],
            [False, False, True, True, True, True, True],
        )

true_only = h[{2: bh.loc(True)}] # should work
```

I think this should work just like `Regular` axes and `h` should be reducible. But it does not.

```log
ValueError: axis 2 is not reducible
```

PR #31, opened 2020-07-07
Nino/update-bh-0.8.0

LovelyBuggies:
This PR is for #29.


LovelyBuggies:
@henryiii Able to merge.

henryiii:
Density calc is wrong, I have the correct calc in talk and probably the bh docs.

src/hist/core.py
```diff
@@ -66,13 +66,20 @@ def project(self, *args: Union[int, str]):
                 f"Only projections by indices and names are supported for {self.__class__.__name__}"
             )
 
-    def plot(self, *args, **kwargs,) -> Union[Plot1D_RetType, Plot2D_RetType]:
+    def density(self):
+        """
+        Density Histogram.
+        """
+
+        return self / self.sum()  # ToDo: maybe should not be filled
```
henryiii, line 74: See https://boost-histogram.readthedocs.io/en/latest/usage/quickstart.html#computing-with-histograms

Also, overflow bins need to be removed. Let's drop this for now and try it in a new PR.

LovelyBuggies: Okay, but luckily, when I use `self.view() / self.sum() / np.prod(self.axes.widths, axis=0)`, the flow bins are removed. Not sure why, but it meets your requirement :)

henryiii: That's because `.view(flow=False)` removes the flow bins, but it also returns a NumPy array, so now there are no flow bins, but also no axes, plot functions, etc.

LovelyBuggies: Use the np array to create a new Histogram? Or just leave it to the next PR.

henryiii: Let's work on it in a new PR. We can a) make a density class - but a lot might be duplicated if we are not careful, or b) make a new histogram, removing flow bins, then input the np array above.

LovelyBuggies: Okay, so what else should I do in this PR?

henryiii: I think that's it.

PR #37, opened 2020-07-10
doc: Support RTD and add README

henryiii:
A slightly more functional readme.


henryiii:
Thanks! We have a RTD build now. We need to add some content / notebooks to the docs, and also make the install instructions something for "normal" users - dev install needs to be clearly marked as dev install instructions.

Issue #35, opened 2020-07-10
Original hist replacement


henryiii:
This is the MIT licensed source of the original hist:

```python
#!/usr/bin/env python
__author__ = 'Ron'

import sys
import os
import argparse
from collections import defaultdict


def plot_histogram(x_values, y_values, title=None, fn=None):
    import pylab

    x = range(len(x_values))
    pylab.bar(x, y_values)
    pylab.xticks(x, x_values, horizontalalignment='center', rotation='vertical')
    pylab.subplots_adjust(bottom=0.2)
    if title:
        pylab.title(title)
    if fn:
        pylab.savefig(fn)
    else:
        pylab.show()

def ioctl_GWINSZ(fd):
    try:
        import fcntl, termios, struct, os
        cr = struct.unpack('hh', fcntl.ioctl(fd, termios.TIOCGWINSZ, '1234'))
    except:
        return
    return cr


def get_terminal_size():
    env = os.environ

    cr = ioctl_GWINSZ(0) or ioctl_GWINSZ(1) or ioctl_GWINSZ(2)
    if not cr:
        try:
            fd = os.open(os.ctermid(), os.O_RDONLY)
            cr = ioctl_GWINSZ(fd)
            os.close(fd)
        except:
            pass
    if not cr:
        cr = (env.get('LINES', 25), env.get('COLUMNS', 80))
    return int(cr[1]), int(cr[0])


def calculate_histogram(fd=sys.stdin, verbose=True, max_buckets=20, screen_width=80):
    hist = defaultdict(int)
    for n in fd:
        try:
            num = round(float(n), 5)
        except:
            continue

        hist[num] += 1

    hist_keys = sorted(hist.keys())
    hist_range_min = hist_keys[0]
    hist_range_max = hist_keys[-1]
    hist_range = hist_range_max - hist_range_min

    total_keys = len(hist_keys)
    num_buckets = min(max_buckets, total_keys)

    step_size = (hist_range_max - hist_range_min) / num_buckets

    y_values = [0.0] * num_buckets
    x_values = [x * step_size + hist_range_min for x in xrange(num_buckets)]

    total_size = 0.0

    for key, val in hist.iteritems():
        bucket = int(((key - hist_range_min) / hist_range) * num_buckets)
        if bucket == num_buckets:
            bucket = num_buckets - 1

        y_values[bucket] += val
        total_size += val

    for i in xrange(len(y_values)):
[...]

PR #40, opened 2020-07-11
feat: adding something like the classic hist

henryiii:
Addressing #35. 

Currently it seems that axes without names are broken in Hist. If I do a new `pip install -e .[dev]`, then try, from a command line:

```bash
hist
1
2
3
2
```
Then press control-D, I get:

```python
Traceback (most recent call last):
  File "/Users/henryschreiner/git/scikit-hep/hist/.env/bin/hist", line 33, in <module>
    sys.exit(load_entry_point('hist', 'console_scripts', 'hist')())
  File "/Users/henryschreiner/git/scikit-hep/hist/src/hist/classic_hist.py", line 22, in main
    h = bh.numpy.histogram(values, bins=args.buckets, histogram=hist.Hist)
  File "/Users/henryschreiner/git/scikit-hep/hist/.env/lib/python3.8/site-packages/boost_histogram/numpy.py", line 111, in histogram
    return histogramdd((a,), (bins,), (range,), normed, weights, density, **kwargs)
  File "/Users/henryschreiner/git/scikit-hep/hist/.env/lib/python3.8/site-packages/boost_histogram/numpy.py", line 74, in histogramdd
    hist = cls(*axs, storage=bh_storage).fill(*a, weight=weights, threads=threads)
  File "/Users/henryschreiner/git/scikit-hep/hist/src/hist/core.py", line 38, in __init__
    if ax.name in self.names:
  File "/Users/henryschreiner/git/scikit-hep/hist/src/hist/_internal/axis.py", line 52, in name
    return self.metadata["name"]
TypeError: 'NoneType' object is not subscriptable
```

@LovelyBuggies, could you take a look? This is really just running `bh.numpy.histogram([1,2,3,2], histogram=hist.Hist)`.


henryiii:
Added two tests that should pass, but don't.

henryiii:
Git has trouble tracking changes to the file that was moved and changed at the same time. Normally git is fine if a file is moved, but the move and the change must happen in separate commits and not be squashed.

LovelyBuggies:
> Git has trouble tracking changes to the file that was moved and changed at the same time. Normally git is fine if a file is moved, but the move and the change must happen in separate commits and not be squashed.

Doesn't matter. The main reason that causes the test failed is that you have non-named axes in Hist. But axis names should not be duplicated.

henryiii:
Non-named axes are explicitly allowed in Hist, that's the difference between Hist and NamedHist. Hist allows names but does not require them, and NamedHist forces names.

What we should do is in the check for duplicate names, simply ignore None (or all False-like might be better), so those don't count in the uniqueness check.

LovelyBuggies:
> simply ignore None 

Exactly.

henryiii:
Feel free to add `@pytest.mark.xfail` for tests you plan to fix later. https://docs.pytest.org/en/latest/skipping.html

LovelyBuggies:
> Feel free to add `@pytest.mark.xfail` for tests you plan to fix later. https://docs.pytest.org/en/latest/skipping.html

I have figured it out. But shortcuts' tests are a little cursory. Merge this if the classic hist is alright, and I will open a new PR to expand these tests later.

henryiii:
We should add an integration test. I’ll have to suggest something.

In conda, this should be pretty much automatically available, bin works the same way, though it might be nice to add histoprint to conda if we add Hist to conda.

LovelyBuggies:
I expect we can walk on two legs, that is we can dev and users are allowed by both conda and venv :)

PR #43, opened 2020-07-11
fix: Title should work over name

henryiii:
@LovelyBuggies, later we should make title a property that returns name if it is unset; that way plotting APIs (including our own) can just use `.title` and be assured that it is the best title available. Maybe `.title` can even return "axis 1" if neither is set. :)

I've copied the precommit file in from boost-histogram, and make the fixes flake8 recommended.


henryiii:
This should make the CI a little more stable, too (setting a seed).


@LovelyBuggies, look over the changes, but I'm working quickly and adding several things per PR (which is not idea, but I'm in a hurry :( )

LovelyBuggies:
> @LovelyBuggies, later we should make title a property that returns name if it is unset; that way plotting APIs (including our own) can just use `.title` and be assured that it is the best title available. Maybe `.title` can even return "axis 1" if neither is set. :)

One way to do that is to assign name or "axis 0" when initializing Hist() if the title is not provided. If the user forgets to add title, it can still use `h.title = "newTitle"` to reset it.

henryiii:
Probably something like that, yes. I'm working on upgrading boost-histogram to 0.9.0, and fixing a few things; expect axis.py to change quite a bit. Should have a PR soon.

PR #45, opened 2020-07-12
feat: update bh, factor out title/name

henryiii:
Adds title if name and title are not set.

@LovelyBuggies take a look, but I'll merge soon, as we need this for @jpivarski's talk. I think it's about the last thing we need in before Monday. I'll make a 2.0 alpha 0 release soon!

### DIFF
--- a/src/hist/axis.py
+++ b/src/hist/axis.py
@@ -1,13 +1,57 @@
 # -*- coding: utf-8 -*-
 from typing import Dict, List, Union
 
+import sys
 import re
 import boost_histogram.axis as bha
 import hist.utils
+from typing import Any
+
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
+
+
+class AxisProtocol(Protocol):
+    metadata: Any
+    name: str
+    title: str
+
+
+class AxesMixin:
+    __slots__ = ()
+
+    @property
+    def name(self: AxisProtocol) -> str:
+        """
+        Get or set the name for the Regular axis
+        """
+        return self.metadata.get("name", "") if isinstance(self.metadata, dict) else ""
+
+    @name.setter
+    def name(self: AxisProtocol, value: str) -> None:
+        self.metadata["name"] = value
+
+    @property
+    def title(self: AxisProtocol) -> str:
+        """
+        Get or set the title for the Regular axis
+        """
+        title = (
+            self.metadata.get("title", "") if isinstance(self.metadata, dict) else ""
+        )
+        return title or self.name
+
+    @title.setter
+    def title(self: AxisProtocol, value: str) -> None:
+        self.metadata["title"] = value
 
 
 @hist.utils.set_family(hist.utils.HIST_FAMILY)
-class Regular(bha.Regular):
+class Regular(bha.Regular, AxesMixin):
+    __slots__ = ()
+
     def __init__(
         self,
         bins: int,
@@ -30,7 +74,7 @@ class Regular(bha.Regular):
         else:
             raise Exception("Name should be a valid Python identifier")
         metadata["title"] = title
-        super(bha.Regular, self).__init__(
+        super().__init__(
             bins,
             start,
             stop,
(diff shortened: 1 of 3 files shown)


src/hist/axis.py
```diff
@@ -30,7 +70,7 @@ def __init__(
         else:
             raise Exception("Name should be a valid Python identifier")
         metadata["title"] = title
-        super(bha.Regular, self).__init__(
```
henryiii, line 33: In Python 3, I would always use `super()` by itself. It's easy to make a mistake. Such as here - `super` takes the *current* class, not the parent. In boost-histogram 0.8.0, I had these double-wrapped, so you didn't notice the "one up", expect for the `__repr__` not working. But in 0.9.0, there is no double-wrapping, so this was grabbing the wrong class. You could fix it by using `super(Regular, self)`, but better yet, just use `super()`.

src/hist/axis.py
```diff
@@ -42,35 +82,9 @@ def __init__(
...
-        return self.metadata["title"] if self.metadata else ""
-
-    @title.setter
-    def title(self, value: str) -> None:
-        self.metadata["title"] = value
-
-
-class Boolean(bha.Regular):  # ToDo: should be derived from bha.Boolean
+@hist.utils.set_family(hist.utils.HIST_FAMILY)
+class Boolean(bha.Boolean, AxesMixin):
```
henryiii, line 87: We can finally use this in 0.9.0 :)

src/hist/core.py
```diff
@@ -45,24 +45,32 @@ def __call__(self, *args, **kwargs):
...
             self._hist = None
-            self._storage_proxy = None
+            self._ax = []
+            super().__init__(*args, **kwargs)
+            valid_names = [ax.name for ax in self.axes if ax.name]
+            if len(valid_names) != len(set(valid_names)):
+                raise KeyError(
+                    f"{self.__class__.__name__} instance cannot contain axes with duplicated names"
+                )
+            for i, ax in enumerate(self.axes):
```
henryiii, line 67: This adds default titles if both name and title are unset. Actually, now that I think about it, we probably only need `if ax.title`, since we now have the title draw from the name if unset...

src/hist/core.py
```diff
@@ -45,24 +45,32 @@ def __call__(self, *args, **kwargs):
 
 @hist.utils.set_family(hist.utils.HIST_FAMILY)
 class BaseHist(bh.Histogram):
+    __slots__ = ("_ax", "_storage_proxy")
+
     def __init__(self, *args, **kwargs):
```
henryiii, line 50: The old ordering that I had here was incorrect, and could trip up the `__getattribute__`. By setting the members at the top here, we avoid errors in `__getattribute__`.

henryiii:
Will merge and push an alpha release when CI passes.

src/hist/axis.py
LovelyBuggies: Okay.

src/hist/core.py
```diff
@@ -45,24 +45,33 @@ def __call__(self, *args, **kwargs):
...
+            self._ax = []
+            super().__init__(*args, **kwargs)
+            valid_names = [ax.name for ax in self.axes if ax.name]
+            if len(valid_names) != len(set(valid_names)):
+                raise KeyError(
+                    f"{self.__class__.__name__} instance cannot contain axes with duplicated names"
+                )
+            for i, ax in enumerate(self.axes):
+                # title will return name if title is not set, so this is safe
+                if not ax.title:
```
LovelyBuggies, line 69: why not

```python
ax.title = ax.title or ax.name or f"Axis {i}"
```

henryiii: Because if `metadata["title"]` is blank, then `.title` returns `metadata["name"]`, so if `.title` is blank, `.name` must be blank.

henryiii: Your way would work if we want to statically set the title always, and leave the logic out of the getters for `.title`. I _think_ I prefer the current method, because it doesn't add a title when there is a name, it just returns the name as the title when asked.

henryiii: We need to have a test to make sure adding the same axes to two Histograms property makes a copy, rather than causing a battle as to who gets to set the title.

LovelyBuggies: https://github.com/scikit-hep/hist/blob/7e76c3f0dd7b4014d5a32f1c048d200bc204eb77/src/hist/core.py#L68

Sorry, I skipped this line 😄

LovelyBuggies: > We need to have a test to make sure adding the same axes to two Histograms property makes a copy, rather than causing a battle as to who gets to set the title.

Better open in a new issue. Let me do it!

PR #48, opened 2020-07-12
refactor: rm redundants

LovelyBuggies:


LovelyBuggies:
@henryiii Reducible problem still exists, could you take a look?

LovelyBuggies:
Consider that ax.titles return names if ax.titles is None, I think these parts could be redundant.

PR #49, opened 2020-07-12
pref: aviod mpl 20 figs warning

LovelyBuggies:


src/hist/core.py
```diff
@@ -539,6 +539,8 @@ def plot_pull(
         fig.add_axes(ax)
         fig.add_axes(pull_ax)
 
+        plt.close(fig)
```
henryiii, line 542: You have to close in the test, not in the source - otherwise, no one will ever be able to use the figure!

henryiii:
We could add a mpl fixture to help with the testing. I've not done much testing with mpl before, will probably look around (like at the mpl source, maybe) to see how it is often done.

src/hist/core.py
LovelyBuggies: yep, I did this because the plt.close did not work in test files so I came up with this idea. Okay, I will TAL at mpl to see how it works, too.

henryiii: No, you are right. This should have fixed the problem, even though it's not a final solution. Sorry! Will have to play with ideas here, then.

henryiii: Though you could try adding this to all ~4 functions, just to make sure it's not the other 3 that are overloading the open plots count.

LovelyBuggies: > Though you could try adding this to all ~4 functions, just to make sure it's not the other 3 that are overloading the open plots count.

One warning lose. Good starting point. But still 35 warnings remained.

LovelyBuggies:
It's headache that the `plt.close()` does not treat this headache only if I add it to:

```python
        """
        Default Figure: construct the figure and axes
        """
        if fig is None:
            fig = plt.figure(figsize=(8, 8))
            grid = fig.add_gridspec(4, 4, hspace=0, wspace=0)
            plt.close(fig)
```

- Adding it before return - doesn't work.
- Closing fig in test files - doesn't work.

For now, I have to mute this warning, as this problem is harmless.

```diff
+ import matplotlib as plt
+ plt.rcParams.update({'figure.max_open_warning': 0})
```

src/hist/basehist.py
```diff
@@ -187,6 +189,8 @@ def plot1d(
 
         fig.add_axes(ax)
 
+        plt.close(fig)
```
henryiii, line 192: These should not be here!

src/hist/basehist.py
```diff
@@ -13,6 +13,8 @@
 from .axis import Regular
 from .axestuple import NamedAxesTuple
 
+plt.rcParams.update({"figure.max_open_warning": 0})
```
henryiii, line 16: Let's not add this quite yet. I'll see if I can find anything to improve this. If not, then we add this.

PR #50, opened 2020-07-12
feat: NamedAxesTuple

### LINKED ISSUE #36: AxesTuple
AxesTuple needs to be extended to support name-based access. We might need a change in boost-histogram to make sure we can override this class everywhere.

`h.axes["x"]` should be the axes with the name "x".

henryiii:
Closes #36.

### DIFF
--- a/src/hist/axestuple.py
+++ b/src/hist/axestuple.py
@@ -31,25 +31,16 @@ class NamedAxesTuple(AxesTuple):
         return super().__getitem__(item)
 
     @property
-    def names(self) -> Tuple[str]:
+    def name(self) -> Tuple[str]:
         """
         The names of the axes. May be empty strings.
         """
         return tuple(ax.name for ax in self)  # type: ignore
 
     @property
-    def titles(self) -> Tuple[str]:
+    def title(self) -> Tuple[str]:
         """
         The titles of the axes. Defaults to name if title not given, or Axis N
         if neither was given.
         """
         return tuple(ax.title for ax in self)  # type: ignore
-
-    # Missing in boost-histogram 0.9.0
-    @property
-    def sizes(self) -> Tuple[int]:
-        """
-        The sizes of the axes.
-        """
-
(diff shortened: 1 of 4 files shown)


LovelyBuggies:
Maybe add some basic tests?

henryiii:
Oops, forgot to commit the new file, thanks!

henryiii:
No, I added them to `tests/test_general.py`, they should be there.

LovelyBuggies:
classic_hist is using axis.size as buckets, bh is using axis.size as bins. I think we need to unify them.

henryiii:
classic_hist is just replicating the old interface of "hist". We can ignore it for now; eventually we might want to just wrap histoprint's CLI (which is much better than ours) with some way to make image files from matplotlib if you pass an output (which is the only new thing we add).

bins and size are different. bins is a keyword argument for some (but not all) the axes - Integer and Category axes do not have a way to set `bins=`. `.size` is the number of bins (and I just picked this up from boost-histogram, I forgot the history as to why `ax.size` exists instead of just `len(ax)`.

I need to remove `hist.names` in favor of `hist.axes.names`, which is more consistent. Will do that now.

henryiii:
Only matrices are plural. Otherwise, it's the singular and exactly matches the axis property. I forgot about that.

henryiii:
So this should be name and title. `hist.axes.name`, `hist.axes.title`. A little ugly, I wonder if they should all be made plural? Will be a question for boost-histogram.

henryiii:
Singular is very ugly. I'll keep these plural and ask in boost-histogram about renaming size -> sizes, etc.

henryiii:
Changed my mind again - will follow bh naming schema (which I realize never change the method/property name, "centers" starts out plural. Asked about this on the boost-histogram issues.

LovelyBuggies:
Amazing feature, pretty hand-on for users!

PR #52, opened 2020-07-12
test: complex index

LovelyBuggies:


tests/test_general.py
```diff
@@ -639,6 +639,53 @@ def pdf(x, a=1 / np.sqrt(2 * np.pi), x0=0, sigma=1, offset=0):
...
+        axis.Integer(0, 10, name="Int"),
+    ).fill(
+        np.random.normal(size=10),
+        np.random.uniform(size=10),
+        ["hi"] * 8 + ["hello"] * 2,
+        [True] * 6 + [False] * 4,
+        np.ones(10),
+    )
+
+    assert (
```
henryiii, line 661: These look great! One question, what are these equal to? Then you should write:

```python
assert ...  == <number>
```

for each item. Test reporting on failures will be *much* better, it's easier for us to read what is happening, and if fill or maybe indexing is broken, this will all pass since 0 == 0 == 0 == 0.

LovelyBuggies: Because they represent the same bin. Note that I used the random so the `<number>` could not be found.

LovelyBuggies: Okay, I changed it to one statement per line.

henryiii: If this is often 0 (which it will be, if you put 10 items into a large histogram), then this really isn't checking anything about the indexing. You can set the seed, in which case you get the same thing each time, or you can avoid random - having any random behavior in tests is bad, generally, and it makes it nearly impossible to debug or catch a problem when it occurs.

PR #53, opened 2020-07-12
test: density

LovelyBuggies:


tests/test_base.py
```diff
@@ -692,3 +692,13 @@ def test_base_index_access():
...
+
+
+def test_base_density():
+    """
+        Test base density -- whether BaseHist can be accessed by index.
+    """
+
+    for data in range(10, 20, 10):
+        h = BaseHist(axis.Regular(10, -3, 3, name="x")).fill(np.random.randn(data))
+        assert round(sum(h.density()), 2) == round(10 / 6, 2)
```
henryiii, line 704: Please use

```python
from pytest import approx

...

assert sum(h.density()) == approx(10 / 6)
```
You can use the keyword parameters to make this more or less approximate.

https://docs.pytest.org/en/stable/reference.html#pytest-approx

(same for the other tests)

LovelyBuggies: Okay.

PR #54, opened 2020-07-12
test: proxy tests

LovelyBuggies:


LovelyBuggies:
There's a bug for NamedHist proxy. And I forget to add proxy for other typed axes like Variables.

LovelyBuggies:
> There's a bug for NamedHist proxy. And I forget to add proxy for other typed axes like Variables.

Do we need to design a decorator and getattribute exclusive for NamedHist?

https://github.com/scikit-hep/hist/blob/1d8643a613132ff3421d0e65841a655f944abd37/tests/test_named.py#L802-L809

LovelyBuggies:
The "fix: make proxy work for NamedHist" should be in the next PR. Let's merge it first if tests have no problem.

tests/test_general.py
```diff
@@ -685,50 +686,41 @@ def test_general_index_access():
...
+        .fill([0.5, 0.5], [-0.2, 0.6])
+    )
 
-def test_histogram_quick_constrution():
-    h = Hist.Regular(10, 0, 1, name="x").Regular(10, -1, 1, name="y")
-    h.fill([0.5, 0.5], [-0.2, 0.6])
-    assert h[0.5j, -0.2j] == 1
-
-
-def test_histogram_unnamed_axes():
```
henryiii, line 711: Where did this test go?

tests/test_base.py
```diff
@@ -694,6 +694,32 @@ def test_base_index_access():
...
+    assert h[0.5j] == 2
+
+    h = (
+        BaseHist()
+        .Regular(10, 0, 1, name="x")
+        .Regular(10, -1, 1, name="y")
+        .fill([0.5, 0.5], [-0.2, 0.6])
+    )
+
+    assert h[0.5j, -0.2j] == h[bh.loc(0.5), bh.loc(0.6)] == 1
```
henryiii, line 711: Please write asserts 1 per line, with the "answer" on each line. That way, reporting doesn't spit way to much text out when there's a failure. And it's easier to read.

```suggestion
    assert h[0.5j, -0.2j] == 1
    assert h[bh.loc(0.5), bh.loc(0.6)] == 1
```

tests/test_general.py
```diff
@@ -685,50 +686,41 @@ def test_general_index_access():
...
-    h[0j] == 1
-    h[2.1j] == 1
+    # add axes to existing histogram
+    with pytest.raises(Exception):
+        Hist().Regular(10, 0, 1, name="x").fill([0.5, 0.5]).Regular(10, -1, 1, name="y")
 
+    with pytest.raises(Exception):
+        Hist(axis.Regular(10, 0, 1, name="x")).Regular(10, -1, 1, name="y")
 
-def test_histogram_rebin():
```
henryiii, line 723: You added this one in a previous PR, I think, in an indexing test?

tests/test_named.py
```diff
@@ -800,6 +800,34 @@ def test_named_index_access():
...
+    assert h[0.5j] == 2
+
+    h = (
+        NamedHist()
+        .Regular(10, 0, 1, name="x")
+        .Regular(10, -1, 1, name="y")
+        .fill(x=[0.5, 0.5], y=[-0.2, 0.6])
+    )
+
+    assert h[0.5j, -0.2j] == h[bh.loc(0.5), bh.loc(0.6)] == 1
```
henryiii, line 819: Same change as above.

tests/test_named.py
```diff
@@ -800,6 +800,34 @@ def test_named_index_access():
...
+
+    # add axes to existing histogram
+    with pytest.raises(Exception):
```
henryiii, line 822: I recommend putting the exact exception here if possible, just to ensure it doesn't change without us knowing.

henryiii:
> The "fix: make proxy work for NamedHist" should be in the next PR.

Tests finding issues is great! That's why we add them.

> Do we need to design a decorator and getattribute exclusive for NamedHist?

Probably, let's not add the rest of the shortcuts until we fix this. We may need to move the "trickery" (`__getattr__`) to Hist from BaseHist, and then add it to NamedHist separately. I think the other parts are general.

tests/test_base.py
LovelyBuggies: Okay.

tests/test_general.py
LovelyBuggies: This should be included above in the initialization part. If not, I'll add it.

tests/test_general.py
LovelyBuggies: Yep!

tests/test_named.py
LovelyBuggies: Okay.

tests/test_named.py
LovelyBuggies: Yes, I agree. This will be a huge task, and I'll find some time to specialize in doing this.

tests/test_general.py
LovelyBuggies: Move it above.

tests/test_named.py
LovelyBuggies: add ToDo item - https://github.com/scikit-hep/hist/pull/54/commits/3f23a6537befec4b8a2c0da00f0158aa196d48be

LovelyBuggies:
@henryiii Able to merge now. Solved all issues related to tests.

henryiii:
> The "fix: make proxy work for NamedHist" should be in the next PR

Feel free to open an issue and assign it to me if you want to.

PR #63, opened 2020-07-13
fix: closes #58

### LINKED ISSUE #58: Fix Hist and NamedHist
* Does named indexing work on all Hist's? I don’t see the code for it in BaseHist, only in NamedHist. Something like this: `h[{“x”:2}]` should be converted into `h[{0:2}]` assuming that axes 0 has name `"x"`. 
* NamedHist should simply verify that the index item is a dict, and that it has only string keys. Then it should just call BaseHist's indexing.
* NamedHist's fill should simply be: `def fill(self, **kwargs): return super().fill(**kwargs)` (maybe with a few more keyword arguments listed to be nice to the user's inspection tools). BaseHist should be able to fill via kwargs or via position.

Overall, NamedHist should be pretty short, it is just taking away functionally (position-based access) from BaseHist. I'm not sure how strict we want to be - we can also have an `OnlyNamedAxesTuple` too.

Also we should make sure you can cast between Hist and NamedHist. `Hist(NamedHist(...))`, etc. I can do that one.

LovelyBuggies:


LovelyBuggies:
And modify the coding style based on the feat of AxisTuple.

LovelyBuggies:
There is an odd thing in this PR. I add some xfail markers. Please look at Look.ipynb, which is a code snippet I directly copied from test_named.py. Not sure why the notebook run well but the pytest failed.

src/hist/basehist.py
```diff
@@ -140,6 +140,17 @@ def __getattribute__(self, item):
 
         return object.__getattribute__(self, item)
 
+    def _name_to_index(self, name: str) -> int:
```
henryiii, line 143: Good refactor, I was going to suggest this. :)

src/hist/basehist.py
```diff
@@ -150,20 +161,140 @@ def project(self, *args: Union[int, str]):
...
             for name in args:
-                for index, axis in enumerate(self.axes):
-                    if name == axis.name:
-                        indices += (index,)
-                        break
-                else:
-                    raise ValueError("The axis names could not be found")
+                if isinstance(
+                    name, str
+                ):  # ToDo: redundant line - mypy cannot realized it
```
henryiii, line 166: Then use:

```python
assert isinstance(name, str) # for MyPy
```

here. This will make mypy believe you.

src/hist/basehist.py
```diff
@@ -150,20 +161,140 @@ def project(self, *args: Union[int, str]):
...
 
             return super().project(*indices)
 
         else:
             raise TypeError(
-                f"Only projections by indices and names are supported for {self.__class__.__name__}"
+                f"Only projection by indices or names is supported for {self.__class__.__name__}"
+            )
+
+    def fill(self, *args, **kwargs):
```
henryiii, line 176: Suggestion: try putting keyword only arguments sample, weight, and thread there, then you don't have to worry about them in the dict, and the signature is nicer for users.

```suggestion
    def fill(self, *args, weight=None, sample=None, thread: Optional[int] =None, **kwargs):
```

src/hist/basehist.py
```diff
@@ -150,20 +161,140 @@ def project(self, *args: Union[int, str]):
...
+            return x
+
+    def __getitem__(self, index):
+        """
+            Get histogram item.
+        """
+
+        if isinstance(index, dict):
+
+            if all(isinstance(k, int) for k in index.keys()):
```
henryiii, line 247: We should allow any mix, no reason to be extra restrictive with `all`. Just loop over the keys, if string convert to int, if int just use it, and if anything else, throw the TypeError. The only extra check I'd do would be to make sure we don't add the same int twice, as `{'x': ..., 0: ...}` is valid but should actually throw an error since it's listing the same axes twice.

src/hist/basehist.py
```diff
@@ -150,20 +161,140 @@ def project(self, *args: Union[int, str]):
...
+            index = (index,)
+
+        return super().__getitem__(tuple(self._loc_shortcut(v) for v in index))
+
+    def __setitem__(self, index, value):
+        """
+            Set histogram item.
+        """
+
+        if isinstance(index, dict):
```
henryiii, line 274: You can factor out the index transform logic into a hidden function if you want, it should be the same for these two functions (set/get).

src/hist/namedhist.py
```diff
@@ -36,38 +35,24 @@ def fill(self, *args, **kwargs):
...
 
         if isinstance(index, dict):
-            k = list(index.keys())
-            if isinstance(k[0], str):
-                for key in k:
-                    for idx, axis in enumerate(self.axes):
-                        if key == axis.name:
-                            index[idx] = index.pop(key)
-                            break
+            if all(isinstance(k, int) for k in index.keys()):
```
henryiii, line 52: You mean `any`, not `all`, right? Or you can use `not all(isinstance(k, str)...`

src/hist/basehist.py
henryiii: Or you can move the check down here, instead of at the top. That way, you don't loop over everything twice (as `all`/`any` do a loop inside).

src/hist/basehist.py
henryiii: (And obviously pass them on to the fill's below)

henryiii:
Looking great, though! :)

src/hist/namedhist.py
LovelyBuggies: Yep! Should be any.

src/hist/basehist.py
LovelyBuggies: Yep, I have thought about this previously. But it's a little bit tricky, as `return super().__set/getittem__` are nested in the logic.

src/hist/basehist.py
LovelyBuggies: > We should allow any mix

I have thought about this. But don't you think it's odd to use a mixed style?

src/hist/basehist.py
henryiii: Once you rewrite this to simply transform "index" with 0 or more string into an "index" with only ints, then it will be be disconnected from the logic and will be easy to factor out.

src/hist/basehist.py
henryiii: Yes, but it's not something we should enforce. If it makes sense (ie, axis numbers and axes names are not overlapping), we should allow it. We should have an error if the same number is added "twice" (once as a name, once as a number), though.

henryiii: We probably will not show mixed styles in the examples, for example ;)

henryiii: After you do this, I can look into the xfails.

LovelyBuggies: And eigenvalue error is becoming more and more rampant.

henryiii: We will solve it properly in the near future.

LovelyBuggies: @henryiii Okay, You can have a look. I did not add more codes on duplicate access (access the same axis), but I believe underlying BH would throw an error if we do that. Feel free to add a test related to this issue :)

src/hist/basehist.py
```diff
@@ -155,54 +155,29 @@ def project(self, *args: Union[int, str]):
...
-                else:
-                    indices += (key,)
-
-            return super().project(*indices)
-
-        else:
-            raise TypeError(
-                f"Only projection by indices and names is supported for {self.__class__.__name__}"
-            )
+        int_args = [self._name_to_index(a) if isinstance(a, str) else a for a in args]
```
henryiii, line 158: Just rebuild this, replacing string names with ints. The underlying `.project()` will handle repeated indices the same way manually repeated indices are handled (I don't really even know what `project(0, 0)` does, the C++ handles this)

src/hist/basehist.py
```diff
@@ -155,54 +155,29 @@ def project(self, *args: Union[int, str]):
...
         """
-            Insert data into the histogram using names and indices return \
-            a Hist object.
+        Insert data into the histogram using names and indices, return
+        a Hist object.
         """
 
-        if len(args) and not len(kwargs):
-            return super().fill(*args)
+        data_dict = {
```
henryiii, line 169: Same idea, just process the list once - much faster than multiple loops. Remember any/all do loops too.

src/hist/basehist.py
```diff
@@ -155,54 +155,29 @@ def project(self, *args: Union[int, str]):
...
 
-        elif (
-            isinstance(kwargs, dict)
-            and all(isinstance(k, str) for k in kwargs.keys())
-            and not len(args)
-        ):
-            indices: dict = {}
-            for n, v in kwargs.items():
-                indices[self._name_to_index(n)] = v
+        if set(data_dict) != set(range(len(args), self.ndim)):
```
henryiii, line 174: Might as well support mixed positional and keywords, if anyone wants to try it. This makes sure all remaining axes are accounted for.

src/hist/basehist.py
```diff
@@ -155,54 +155,29 @@ def project(self, *args: Union[int, str]):
...
+            raise TypeError("All axes must be accounted for in fill")
 
-            lst = sorted(indices.items(), key=lambda item: item[0])
-            nd = np.asarray(lst, dtype=object)
-            data = nd.ravel()[1::2]
-            return super().fill(*data)
+        data = (data_dict[i] for i in range(len(args), self.ndim))
```
henryiii, line 177: This builds them in order.

src/hist/basehist.py
```diff
@@ -155,54 +155,29 @@ def project(self, *args: Union[int, str]):
...
 
-        else:
-            raise TypeError(
-                f"Only dict with keys of int and str is supported for {self.__class__.__name__}"
-            )
+        total_data = tuple(args) + tuple(data)  # Python 2 can't unpack twice
```
henryiii, line 179: Actually, I have no idea why I made this Python 2 complaint. Habit, I guess. Feel free to turn back into the simpler:

```python
return super().fill(*args, *data, weight=weight, sample=sample, threads=threads)
```

src/hist/basehist.py
```diff
@@ -284,34 +284,24 @@ def plot1d(
         if self.ndim != 1:
             raise TypeError("Only 1D-histogram has plot1d")
 
-        """
-        Default Figure: construct the figure and axes
-        """
+        # Default Figure: construct the figure and axes
```
henryiii, line 287: Let's use normal comment blocks, and only use strings at the top of functions, where they turn into docstrings.

src/hist/basehist.py
```diff
@@ -327,27 +317,29 @@ def plot2d(
...
             raise TypeError("Only 2D-histogram has plot2d")
 
-        """
-        Default Figure: construct the figure and axes
-        """
-        if fig is None:
-            fig = plt.figure(figsize=(8, 8))
-            grid = fig.add_gridspec(4, 4, hspace=0, wspace=0)
+        # Default Figure: construct the figure and axes
+        if fig is not None and ax is not None:
```
henryiii, line 321: Please take a look at this fig/ax handling, the other functions need it too. If you pass fig and ax, the ax must be in the figure. If you pass one, the other is created or accessed.

src/hist/basehist.py
```diff
@@ -327,27 +317,29 @@ def plot2d(
...
+            fig = ax.figure
+
+        # Plot: plot the 2d-histogram
         X, Y = self.axes.edges
         ax.pcolormesh(X.T, Y.T, self.view().T, **kwargs)
 
         ax.set_xlabel(self.axes[0].title)
         ax.set_ylabel(self.axes[1].title)
 
-        fig.add_axes(ax)
```
henryiii, line 261: I don't think you ever need this. We don't support mis-matched fig/axes.

src/hist/basehist.py
LovelyBuggies: https://github.com/scikit-hep/hist/blob/6fb66d385bbc57c5dea7e2f9607df416242f9ad1/tests/test_general.py#L330-L331

PR #68, opened 2020-07-16
Feat/storage

### LINKED ISSUE #39: Export parts of boost-histogram
We should provide parts of boost-histogram that don't need changing; specifically the numpy module should be provided. The histogram argument is designed to be able to take a Hist class instead of a Histogram class (probably will break with NamedHist, though, as it's not going to add names; but I think that's okay, because NumPy has no method to add names).

henryiii:
* Adds missing files, closes #39 
* Adds storage shortcuts
* Fixes open figure warning, closes #49

### DIFF
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -5,11 +5,15 @@ import matplotlib
 import matplotlib.pyplot as plt
 import matplotlib.patches as patches
 from matplotlib import transforms
+import warnings
 from scipy.optimize import curve_fit
 from uncertainties import correlated_values, unumpy
 from typing import Callable, Optional, Tuple, Union, List, Any
 
 import hist.utils
+import hist.storage
+from hist.storage import Storage
+
 from .axis import Regular, Boolean, Variable, Integer, IntCategory, StrCategory
 from .axestuple import NamedAxesTuple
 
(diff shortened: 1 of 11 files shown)


src/hist/basehist.py
```diff
@@ -5,11 +5,15 @@
...
 import matplotlib.patches as patches
 from matplotlib import transforms
+import warnings
 from scipy.optimize import curve_fit
 from uncertainties import correlated_values, unumpy
 from typing import Callable, Optional, Tuple, Union, List, Any
 
 import hist.utils
+import hist.storage
+from hist.storage import Storage
```
henryiii, line 15: It's nice to leave full names for things unless they are common or part of a complex expression, or are Types.

src/hist/basehist.py
```diff
@@ -48,19 +52,28 @@ def __call__(self, *args, **kwargs):
...
         self._ax: List[bh.axis.Axis] = []
         self._hist: Any = None
-        self._storage_proxy: Any = None
+        self._storage_proxy: Optional[Storage] = None
         self.axes: NamedAxesTuple
 
         if len(args):
             self._hist = None
             self._ax = []
+            kwargs = {}  # Only needed for boost-histogram < 0.10.0
```
henryiii, line 66: I've fixed this requirement upstream, we should be able to do `metadata=metadata, storage=storage` soon.

tests/test_base.py
```diff
@@ -607,13 +610,17 @@ def pdf(x, a=1 / np.sqrt(2 * np.pi), x0=0, sigma=1, offset=0):
...
+    plt.close("all")
+
     # wrong kwargs names
     with pytest.raises(Exception):
         h.plot_pull(pdf, abc="crimson", xyz="crimson")
 
     with pytest.raises(Exception):
         h.plot_pull(pdf, ecolor="crimson", mfc="crimson")
 
+    plt.close("all")
```
henryiii, line 622: The exceptions still open figures, that's why you were not able to stop this from having 20+ plots. We really should make sure exceptions don't open plots if possible!

tests/test_base.py
```diff
@@ -690,6 +699,23 @@ def test_base_index_access():
         h[0:10:20j, 0:5:10j, "hello", False, 5]
 
 
+def test_storage_proxy_int():
```
henryiii, line 702: You should have one test function per case. If you really want to bundle similar tests, use a class, but really, your naming scheme should do the bundling - if you want to run only `test_storage_proxy`, then you use `-k test_storage_proxy` and that will run this "group".

src/hist/basehist.py
```diff
@@ -87,6 +98,42 @@ def Reg(self, *args, **kwargs):
...
 
+    @always_normal_method
+    def Sqrt(self, *args, **kwargs):
+        if self._hist:
+            raise RuntimeError("Cannot add an axis to an existing histogram")
+        self._ax.append(Regular(*args, transform=hist.axis.transform.sqrt, **kwargs))
+        return self
+
+    @always_normal_method
+    def Log(self, *args, **kwargs):
```
henryiii, line 109: For all these functions we are wrapping, we should look up and use the real Python 3 signature from boost-histogram's decorators. That way, tab completion and docs are better.

PR #71, opened 2020-07-16
feat: boost-histogram 0.10.0

henryiii:
Update to bh 0.10.0

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,11 +44,11 @@ package_dir =
   =src
 install_requires =
   numpy >=1.13.3
-  boost-histogram ==0.9.0
+  boost-histogram ==0.10
   matplotlib >=2.0
-  scipy >=1.4.1
+  scipy >=1.4
   uncertainties >=3
-  histoprint >= 1.4.0
+  histoprint >= 1.4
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -44,11 +44,11 @@ package_dir =
   =src
 install_requires =
   numpy >=1.13.3
-  boost-histogram ==0.9.0
+  boost-histogram ==0.10
```
henryiii, line 47: We now are allowing patch releases here.

src/hist/basehist.py
```diff
@@ -375,7 +374,9 @@ def density(self):
         """
         # ToDo: maybe should not be filled
         # ToDo: flow should be removed
-        return self.view() / self.sum() / np.prod(self.axes.widths, axis=0)
```
henryiii, line 378: `reduce` is a better way to write this - and the `prod` is also broken in 0.10.0. I have a PR that fixes it and removes the warning seen when using the old method and NumPy 1.19.

henryiii:
@LovelyBuggies We can make a beta release when you are ready, but I'm going to do a quick alpha release so I can work on my talk.

PR #75, opened 2020-07-17
feat: plot 1D, 2D return artists

henryiii:
My fix for just 1D/2D.

### DIFF
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -57,7 +57,7 @@ class BaseHist(bh.Histogram):
             Initialize BaseHist object. Axis params can contain the names.
         """
         # TODO: Make a base class type Axis for Hist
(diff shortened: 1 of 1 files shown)


src/hist/basehist.py
```diff
@@ -378,12 +378,14 @@ def density(self):
             self.sum() * functools.reduce(operator.mul, self.axes.widths)
         )
 
-    def plot(self, *args, **kwargs) -> Union[Plot1D_RetType, Plot2D_RetType]:
+    def plot(
+        self, *args, **kwargs
+    ) -> Union[matplotlib.lines.Line2D, matplotlib.collections.QuadMesh]:
```
henryiii, line 383: This also works:

```suggestion
    ) -> matplotlib.artist.Artist:
```

PR #76, opened 2020-07-17
fix: 1d and 2d plot & title-> label

LovelyBuggies:
- [x] fix: return type of plotting family 
- [x] fix :fig and ax pass in
- [x] docs: make a slight change to allow all names


LovelyBuggies:
Two issues before merge @henryiii :

- [ ] Should I give ax_res value everytime it draws a different plot (errorbar then plot then show_diff, etc.)
- [ ] There are many assert in the basehist.py to teach mypy the data type. Better if we can remove them and pass.

henryiii:
> 2.0.0 beta1

Let's rename title to label before releasing a beta.

src/hist/basehist.py
```diff
@@ -468,38 +459,57 @@ def plot2d(
 
     def plot2d_full(
         self,
+        *,
         fig: Optional[matplotlib.figure.Figure] = None,
         main_ax: Optional[matplotlib.axes.Axes] = None,
         top_ax: Optional[matplotlib.axes.Axes] = None,
         side_ax: Optional[matplotlib.axes.Axes] = None,
```
henryiii, line 466: I thought this was going to be `ax_dict =`

LovelyBuggies: I have implemented in this way hours ago. But abandoned, think about this:

```python
plot2d_full(fig=fig, ax_dict={"main": main_ax, "top": top_ax, "side": side_ax}, ...)
plot2d_full(fig=fig, main_ax=main_ax, top_ax=top_ax, side_ax=side_ax, ...)
```

Which would be a better shortcut for HEPists? I think the second, as they won't require structured manipulation -- encapsulate to a dict (or list, set ...) -- and only directly pass objects in. But if you want them to be better-structured, it's also reasonable.

henryiii: The problem is that your second example doesn't tell me that I need all 3 at the same time. There's no way in Python's signatures to indicate that one keyword argument will require the others. While the dict clearly tells me it it needs several axes. We can even make a TypedDict and MyPy might help a user add the correct keys.

I don't expect users to often make & pass the dict - I expect 90% of users to leave it empty, and many of the remaining 10% can pass `fig=` and still have the axes created for them.

LovelyBuggies: Oh, just to note that, as for the manipulation for the pass-in arguments, we can handle them in both ways.

henryiii: You can always write:

```python
plot2d_full(fig=fig, ax_dict=dict(main_ax=main_ax, top_ax=top_ax, side_ax=side_ax), ...)
```

henryiii:
>  ax_res value everytime it draws a different plot

The result should always be the same. If there's a way to draw a colorbar directly, don't return it - a user who wants it should do it the proper way in matplotlib.

LovelyBuggies:
> The result should always be the same.

So just assign once is okay?

> Let's rename title to label before releasing a beta.

Figure it out in the next PR.

src/hist/basehist.py
LovelyBuggies: Okay, will change back to it 😁

LovelyBuggies: Do we also need to give ax_dict to plot1d and plot2d (one axis plotting) to keep the uniform of API?

henryiii: No, plotting onto multiple axes is different than the (more common) plot to one axis.

henryiii:
I'm going to manually squash all but the last commit, then rebase & merge. :)

PR #79, opened 2020-07-27
feat: hist svg

LovelyBuggies:


LovelyBuggies:
A few points:
We should try the “table” html view - the axes on the right getting cut off and not being “real” text is irritating; if it was just normal wrapping text in HTML that would be nicer.
Let’s put the label if it exists, and the name if it doesn’t, on the sides like a normal plot would have them for 2D. I know there must be a way to rotate text for the y label.
The ND plot should not change based on the number of dimensions, at least after 5 or so, it should just draw a •••. Is the little histogram on it hard coded?
2D and ND should not check for circular - they are valid regardless of whether the axes are circular or not.

PR #93, opened 2020-09-27
refactor!: quick construct uses new system

### LINKED ISSUE #92: [BUG] operator overloads don't work with the shortcut method.
If you do `repr(Hist.Reg(10,0,1))`, that breaks because dunder methods are looked up on the class, not the object.

henryiii:
Closes #92 

I chose not to include this yet:

```python
Hist.new.Reg(10,0,1).fill(...)
```

since one easy mistake might be:

```python
h = Hist.new.Reg(10,0,1)
h.fill(...)
```

Which is incorrect; though the error message when you try to use "h" or if you use types is good:

```python
h: ConstructProxy = Hist.new.Reg(10,0,1)
```

makes the mistake obvious.

TODO: update docs / notebooks.


henryiii:
PS: we need to refactor the tests a bit (not before 2.0); there is a lot of repeated test code (making changes hard to make); using a fixture to test Hist, NamedHist, and BaseHist all at the same time would me _much_ easier to work with. And they should be unit tests, that is, construction should check to see if it is correctly constructed, and "is usable" should be left to separate usability tests.

PR #94, opened 2020-09-28
fix: fix kwargs in .fill() for WeightedMean&Mean storage

LovelyBuggies:
@henryiii Fix the error of "axis name could not be found" in

```python
h = NamedHist.new.Reg(10, 0, 1, name="x").WeightedMean().fill(x=[0.5, 0.5], weight=[1, 1], sample=[1, 1])
```

and allow TestClass to work (update `.new` proxy tests).

### DIFF
--- a/src/hist/quick_construct.py
+++ b/src/hist/quick_construct.py
@@ -103,7 +103,7 @@ class ConstructProxy(QuickConstruct):
     def Int64(self) -> "BaseHist":
         return self.hist_class(*self.axes, storage=storage.Int64())
 
-    def AutomicInt64(self) -> "BaseHist":
+    def AtomicInt64(self) -> "BaseHist":
         return self.hist_class(*self.axes, storage=storage.AtomicInt64())
 
     def Weight(self) -> "BaseHist":
--- a/src/hist/quick_construct.py
+++ b/src/hist/quick_construct.py
@@ -103,7 +103,7 @@ class ConstructProxy(QuickConstruct):
     def Int64(self) -> "BaseHist":
         return self.hist_class(*self.axes, storage=storage.Int64())
 
-    def AtomicInt64(self) -> "BaseHist":
+    def AutomicInt64(self) -> "BaseHist":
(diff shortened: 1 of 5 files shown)


src/hist/quick_construct.py
```diff
@@ -103,7 +103,7 @@ def Double(self) -> "BaseHist":
     def Int64(self) -> "BaseHist":
         return self.hist_class(*self.axes, storage=storage.Int64())
 
-    def AtomicInt64(self) -> "BaseHist":
+    def AutomicInt64(self) -> "BaseHist":
```
henryiii, line 106:
```suggestion
    def AtomicInt64(self) -> "BaseHist":
```

tests/test_base.py
```diff
@@ -691,40 +691,39 @@ def test_base_index_access():
         h[0:10:20j, 0:5:10j, "hello", False, 5]
 
 
-class test_base_storage_proxy:
+class Test_base_storage_proxy:
```
henryiii, line 694:
```suggestion
class TestBaseStorageProxy:
```
Classes should always be CamelCase. :)

tests/test_general.py
```diff
@@ -697,40 +697,39 @@ def test_general_index_access():
     plt.close("all")
 
 
-class test_general_storage_proxy:
+class Test_general_storage_proxy:
```
henryiii, line 700:
```suggestion
class TestGeneralStorageProxy:
```

tests/test_named.py
```diff
@@ -438,40 +438,39 @@ def test_named_access():
         h[0j, -0j + 2, "hi", True]
 
 
-class test_named_storage_proxy:
+class Test_named_storage_proxy:
```
henryiii, line 441:
```suggestion
class TestNamedStorageProxy:
```

src/hist/quick_construct.py
LovelyBuggies: Okay, sorry for the spelling.

LovelyBuggies:
@henryiii Your commit doesn't cover some places, I made a new one.

PR #95, opened 2020-09-28
Refactor: test cleanup plus plotting optional

henryiii:
Redesigning to make the plotting features optional.


tests/test_general.py
```diff
@@ -8,19 +11,31 @@
...
-# ToDo: specify what error is raised
+
+@pytest.fixture(params=[Hist, BaseHist, NamedHist])
+def named_hist(request):
+    yield request.param
+
+
+@pytest.fixture(params=[Hist, BaseHist])
+def unnamed_hist(request):
+    yield request.param
```
henryiii, line 22: @LovelyBuggies Here's how you would parametrize tests. However, since Hist and BaseHist are identical, and BaseHist is not supposed to be user-facing, we should just test Hist in most cases, unless there's a reason to check BaseHist (case in point: `Hist.new` should return a `Hist`, while `BaseHist.new` should return a BaseHist, so that's a good thing to check).

tests/test_general.py
```diff
@@ -36,71 +51,75 @@ def test_general_init():
...
-    assert Hist(
+    assert unnamed_hist(
         axis.IntCategory(range(-3, 3), name="x"),
         axis.IntCategory(range(-3, 3), name="y"),
     ).fill(np.random.randn(10), np.random.randn(10))
 
-    assert Hist(
+    assert unnamed_hist(
         axis.StrCategory(["F", "T"], name="x"), axis.StrCategory("FT", name="y")
     ).fill(["T", "F", "T"], ["T", "F", "T"])
```
henryiii, line 77: @LovelyBuggies These really aren't unit tests. A unit test would check _just_ construction, then _just_ filling (and would assume construction works, obviously, since it has to construct to fill). Tests should be treated like code; you have to rework them and maintain them occasionally, so keeping repetition down, and keeping the tests focused on one thing at a time is important, just like in code. (Just some tips for the future, still very good test coverage here :) )

henryiii: PS: The reason this matters in this case: if they were really unit tests, I could have the named construction run with the `"named_hist"` fixture, which would put in Hist, BaseHist, and NamedHist; but I can't do that because they make axes with names but then fill without names.

Eventually, we should be able to have all tests broken into pieces; the axis-with-names parts would run with NamedHist and one or both of the Hist's, and then name-only and positional-is-okay tests would have a different or no parametrized fixture.

PR #113, opened 2020-10-29
docs: cleanup, add an example

henryiii:


henryiii:
This is too large. We should render it instead. Closing until/unless we render.

henryiii:
@all-contributors please add @eduardo-rodrigues for code

PR #128, opened 2021-02-15
boost-histogram and histoprint dependencies update

eduardo-rodrigues:
Can this be released as 2.1.0 so that we get a consistent set of packages for the next `scikit-hep` metapackage release, see https://github.com/scikit-hep/scikit-hep/pull/133? Advance thanks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
         python-version:
         - 3.6
         - 3.8
+        - 3.9
     name: Check Python ${{ matrix.python-version }}
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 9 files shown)


henryiii:
Boost-histogram 0.13 out! I'll update and do a bit of testing, add the storage= shortcuts, then release 1.1, should be today or tomorrow.

eduardo-rodrigues:
You mean release 2.1, I'm sure. Sounds great, thanks for the updates.

henryiii:
Yes, that's what I meant. Sorry, my brain is fried (as @HDembinski already knows). :) 👶 

Waiting on a bug in mplhep, matplotlib doesn't like getting an alias kw argument and the insulated name; hoping @andrzejnovak can fix.

henryiii:
Worked around the but for now, will be much nicer when fixed upstream, but that way a release was not delayed. Hist 2.1 is now released. The PlottableHistogram Protocol is now available in boost-histogram 0.13.0, hist 2.1.0, uproot 4, and mplhep 0.2.16! I'm not sure about histoprint.

PR #142, opened 2021-03-09
feat: support data=

henryiii:
See scikit-hep/boost-histogram#421. This now supports setting `data=` on (normal) construction (not supported in QuickConstruct mode).


henryiii:
Not filling out the users guide just yet because it's a pain to start up a jupyterlab instance just to edit docs. I'll probably look into converting the docs into MyST, so that I can edit them in an editor (without futzing with JSON, anyway).

Issue #146, opened 2021-03-12
[BUG] error bars in plot_pull


aminnj:
Should https://github.com/scikit-hep/hist/blob/221d28ffd09fd1dadb87be9306c56146f4095f8b/src/hist/plot.py#L173 have a sqrt? I.e.,
```python
yerr = self.variances() ** 0.5
```
Otherwise, the plot/pulls look to be off for this example.

```python
import numpy as np
from hist import Hist
from uncertainties import unumpy as unp

np.random.seed(42)
hh = Hist.new.Reg(50, -5, 5).Double().fill(np.random.normal(0,1,int(1e5)))

def func(x, constant, mean, sigma):
    exp = unp.exp if constant.dtype == np.dtype("O") else np.exp
    return constant * exp(-((x - mean) ** 2.0) / (2 * sigma ** 2))

hh.plot_pull(func)
```

![image](https://user-images.githubusercontent.com/5760027/111009669-274ad900-8349-11eb-8e61-2add7887ad3b.png)

Changing it wrt master gives me this.

![image](https://user-images.githubusercontent.com/5760027/111009793-709b2880-8349-11eb-9f67-9eecf34ccbf9.png)

Also, I'm wondering if there would be a good reason to not add a little bit more sugar to the fitting. E.g., I could contribute a wrapper around `curve_fit` which recognizes `a=1` as a default and feeds it into `curve_fit`'s `p0`
```python
def f(x, a=1):
    return a*x
```
as in https://github.com/aminnj/yahist/blob/dcbcd055e701775f820e97c95e66599b726530eb/yahist/utils.py#L249.
I also found the switching between `unp` (`uncertainties.numpy`) and `np` in the user callable to be a little unintuitive. Would you consider having imuinuit as a dependency instead of `uncertainties`? Then it's only a couple of extra lines to get the replacement for `curve_fit`'s `pcov` from `iminuit` and resample to get the error band
```python
            vopts = np.random.multivariate_normal(popt, pcov, nsamples)
            sampled_ydata_fine = np.vstack(
                [func(xdata_fine, *vopt).T for vopt in vopts]
            )
```

Let me know if I should move this to a separate issue/discussion :)

henryiii:
Looks like a bug, thanks!

And for your second point, I'd rather like to move to iminuit and drop the SciPy requirement entirely (except for now through mplhep). Happy to have help if you want to. ;)

aminnj:
Great. I'll find some time to make a PR then. :)

[...]

PR #149, opened 2021-03-15
More flexible fitting function, allow likelihood, remove uncertainties dependency

aminnj:
Based on the discussion in https://github.com/scikit-hep/hist/issues/146, I added some features to `plot_pull`. The theme is making fitting more streamlined for exploration.

* Pull `curve_fit` initial guess (`p0`) from default arguments, if they exist
* Allow a string as an alternative to a lambda function (`plot_pull("a+b*x")`)
* Cosmetic change to the band, and embedding fit result into the legend
* Likelihood fit (`plot_pull(..., likelihood=True)`) (chi2 by default, as before)
* Remove `uncertainties.numpy` dependency and construct band by resampling covariance matrix
* Introduce `iminuit` (gets the covariance matrix right, unlike `scipy.optimize` most of the time), but the initial guesses for `iminuit` are seeded from `scipy`

Setup
```python
import numpy as np
from hist import Hist

np.random.seed(42)
hh = Hist.new.Reg(50, -5, 5).Double().fill(np.random.normal(0,1,int(1e5)))
```

Before (including a bug-fix for the `variances` from the above issue):
```python
from uncertainties import unumpy as unp
def func(x, constant, mean, sigma):
    exp = unp.exp if constant.dtype == np.dtype("O") else np.exp
    return constant * exp(-((x - mean) ** 2.0) / (2 * sigma ** 2))

hh.plot_pull(func)
```
image

After:
```python
# as before, but no need for `uncertainties.numpy` as the error band comes
# from resampling the covariance matrix
def func(x, constant, mean, sigma):
    return constant * np.exp(-((x - mean) ** 2.0) / (2 * sigma ** 2))
hh.plot_pull(func)

# `curve_fit` `p0` extracted from defaults, if any
def func(x, constant=80, mean=0., sigma=1.):
    return constant * np.exp(-((x - mean) ** 2.0) / (2 * sigma ** 2))
hh.plot_pull(func)

# strings are allowed to allow for more compactness than a lambda
# x is assumed to be the main variable
hh.plot_pull("constant*np.exp(-(x-mean)**2. / (2*sigma**2))")

# gaussian is a common/special function, so this also works
# reasonable guesses are made for constant/mean/sigma
hh.plot_pull("gaus")
```
image

```python
# chi2 puts `a` around 5, but likelihood puts `a` around 1e3/50 = 20
hh.plot_pull("a+b*x", likelihood=True)
```

image


henryiii:
Can you verify my changes are valid? Could you add a test with pytest-mpl, perhaps?

src/hist/plot.py
```diff
@@ -45,7 +46,7 @@ def _filter_dict(
     }
 
 
-def _expr_to_lambda(expr: str) -> Callable:
+def _expr_to_lambda(expr: str) -> Callable[..., Any]:
```
henryiii, line 49: This is the default for Callable - best to not leave empty Generics.

src/hist/plot.py
```diff
@@ -72,26 +73,29 @@ def _expr_to_lambda(expr: str) -> Callable:
...
+    return eval(lambdastr)  # type: ignore
 
 
 def _curve_fit_wrapper(
-    func: Callable,
-    xdata: ArrayLike,
-    ydata: ArrayLike,
-    yerr: ArrayLike,
+    func: Callable[..., Any],
+    xdata: np.ndarray,
```
henryiii, line 81: ArrayLike can be a simple number, so `xdata[stuff]` was not valid. Either use np.asarray, or just require arrays.

src/hist/plot.py
```diff
@@ -72,26 +73,29 @@ def _expr_to_lambda(expr: str) -> Callable:
...
+    ydata: np.ndarray,
+    yerr: np.ndarray,
     likelihood: bool = False,
-) -> Tuple[ArrayLike, ArrayLike]:
+) -> Tuple[Tuple[float, ...], ArrayLike]:
```
henryiii, line 85: Makes the typing easier, because NumPy's typing is a bit spotty. It doesn't know that `*array` is valid, etc. This is small so was a simple fix.

src/hist/plot.py
```diff
@@ -72,26 +73,29 @@ def _expr_to_lambda(expr: str) -> Callable:
...
 
-    p0 = None
-    if func.__defaults__ and len(func.__defaults__) + 1 == func.__code__.co_argcount:
-        p0 = func.__defaults__
+    params = list(inspect.signature(func).parameters.values())
+    p0 = [
+        1 if arg.default is inspect.Parameter.empty else arg.default
+        for arg in params[1:]
+    ]
+
```
henryiii: I'm using inspect.signature, as it's a more public interface; `__defaults__` (and maybe `__code__`) are untyped, AFAICT (or maybe just not always present on Callables). This also supports partial defaults, while the other was all or nothing.

src/hist/plot.py
```diff
@@ -221,8 +225,8 @@ def plot_pull(
     """
 
     try:
-        from iminuit import Minuit
-        from scipy.optimize import curve_fit
+        from iminuit import Minuit  # noqa: F401
+        from scipy.optimize import curve_fit  # noqa: F401
     except ImportError:
```
henryiii, line 230: This could be a ModuleNotFoundError.

src/hist/plot.py
```diff
@@ -261,20 +265,27 @@ def plot_pull(
         )
     yerr = np.sqrt(variances)
 
-    if type(func) in [str]:
+    if isinstance(func, str):
```
henryiii: Never check the type exactly, always use `isinstance`. It supports subclassing and MyPy type narrowing.

src/hist/plot.py
```diff
@@ -261,20 +265,27 @@ def plot_pull(
...
         if func == "gaus":
             # gaussian with reasonable initial guesses for parameters
-            constant = ydata.max()
+            constant = float(ydata.max())
```
henryiii: Not always fond of NumPy's choices here. Not sure what a `number[Any]` is. So just forcing it to a Python float.

src/hist/plot.py
```diff
@@ -261,20 +265,27 @@ def plot_pull(
...
+            ) -> np.ndarray:
+                return constant * np.exp(-((x - mean) ** 2.0) / (2 * sigma ** 2))  # type: ignore
 
         else:
             func = _expr_to_lambda(func)
 
-    parnames = func.__code__.co_varnames[1:]
+    assert not isinstance(func, str)
+
+    parnames = list(inspect.signature(func).parameters)[1:]
```
henryiii, line 288: Same higher level usage of `inspect.signature` over `.__code__.co_varname`.

henryiii (commented):
Notes on my changes.

src/hist/plot.py
henryiii: Also never make a list for containership testing, use a set, it's faster. `x in {a, b}`.

henryiii:
@all-contributors please add @aminnj for code

henryiii:
Also, would you like to fill in the changelog so we can clear that needs-changelog badge? :)

aminnj:
I appreciate the comments. I haven't yet grasped all the subtleties of the type system...
Is the suggestion to use `pytest-mpl` to make a snapshot of an example plot and use it as a reference image in unit testing? What if the style is changed upstream?

henryiii:
This also need docs, but don't worry about that - I'll be reworking the docs in the near future to make them easier to edit. When I do, I'll add to this part.

There are two ways to test. The easier way is to use `pytest-mpl` to make a snapshot of an example plot and use it as a reference image in unit testing. If the style is changed upstream, we will immediately know and will have to change our test image.

The better way to test is to mock mpl and then verify the sequence of commands to the mpl functions do not change. That also gives much better error messages if something gets changed, rather than just an image diff. But that's much harder to set up. You can see what I did for mplhep here: https://github.com/scikit-hep/mplhep/blob/5effa0b857d7eea93da0299cbbcdc777bd3abaaf/tests/test_mock.py - but as you see, I didn't end up adding it for everything because it's a bit of work for each one.

aminnj:
I added in `pytest-mpl` and seems CI is finally happy with it. Lowered the DPI for the test case made with `pytest tests/test_plot.py --mpl-generate-path=baseline`. By default, it has a very matplotlib 1.x feel.

image

henryiii:
Looks good to me, can you add a changelog entry, then we can merge? If not, let me know, and I'll add one; I usually lightly push contributors to add changelog entries but add them if not; original contributors are usually the best at advertising what they have done. :)

setup.py
```diff
@@ -11,12 +11,14 @@
...
     "scipy >=1.4",
-    "uncertainties >=3",
+    "iminuit >=2",
     "mplhep >=0.2.16",
 ]
 
 extras_require["test"] = [
+    *extras_require["plot"],
     "pytest >=4.6",
+    "pytest-mpl >=0.12",
```
henryiii, line 21: I actually thought we were already using pytest-mpl. Guess I've been staring at mplhep too long. Thanks for adding it!

aminnj:
I added in some test cases that I had forgotten earlier (the `h.plot_pull("gaus")` alias and `likelihood=True`). The changelog entry I added is just a compact summary of the points in the original post of this thread, but let me know if I should include more details.

henryiii (approved):
Looks great, thanks!

eduardo-rodrigues:
Hi @henryiii, @aminnj, I'm following your nice developments. I just wanted to make a comment now that I see `h.plot_pull("gaus")`: I know that ROOT does this, amputating Gauss's name to save a little letter, but I hate that. TBH I see no gain and consider far more important to save the great mathemacian's name. Could we please please do "gauss"?

aminnj:
I personally have no strong attachment to "gaus", so I can offer a +1 to "gauss"

nsmith-:
I was taking a look at https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.html today and noticed it has a `fit` function that does an unbinned max-likelihood fit. I wonder if there's a possibility we could add a `binned_fit` function to scipy? It could take a UHI plottable or numpy array as input.

Issue #148, opened 2021-03-15
[FEATURE] Add ratio plot support to API


matthewfeickert:
As there is already the ability in `hist` to add a ratio plot to the bottom of the main plot, as seen in the [`.plot_pull()` API](https://hist.readthedocs.io/en/latest/user-guide/notebooks/Plots.html#Via-Plot-Pull) it would be great if the ratio plot functionality (which looks super nice and clean in the `plot_pull` examples btw :+1:) could be added to the public API. If people agree this would be useful I'd be willing to contribute a PR.

**Describe the problem, if any, that your feature request is related to**

I don't have a problem per say, but just would like this feature to exist so that I don't have to implement it in a library that is already using `hist`. https://github.com/matthewfeickert/heputils/issues/22

**Describe the feature you'd like**

I'd like to have the ratio plot functionality that exists already in the `plot_pull` API to be available through a public API that would be something like `.plot_ratio`.

**Describe alternatives, if any, you've considered**

None, as this would be up to the developers to decide if this is in scope or not.

henryiii:
Sounds good. Plot_pull could call plot_ratio. Could be easier to upstream to mplhep too.

matthewfeickert:
> Sounds good. Plot_pull could call plot_ratio. 

That makes sense to me

> Could be easier to upstream to mplhep too.

I'll let @henryiii and @andrzejnovak decide where would be the best place for a PR, but at the moment it seems that 

https://github.com/scikit-hep/hist/blob/3827e75fb2587877ca805b30751ab9a8888113de/src/hist/plot.py#L129

is still mostly implemented in `hist` so I'll start work there until someone tells me to pivot.

andrzejnovak:
Ideally, it would exist as a standalone function in mplhep such that one could also use it manually or with uproot TH1s, but since I'm not the one writing it I should shush :D

henryiii:
I'd start by refactoring Hist's function, then we can upstream it and then move to using the new version.

PR #150, opened 2021-03-16
fix: variances was missing sqrt

henryiii:
Fix for bug reported #146 for the next patch version; will get #148 in for the next minor version.

### DIFF
--- a/src/hist/plot.py
+++ b/src/hist/plot.py
@@ -170,7 +170,7 @@ def plot_pull(
 
     # Computation and Fit
     values = self.values()
-    yerr = self.variances() ** 0.5
+    yerr = np.sqrt(self.variances())
 
     # Compute fit values: using func as fit model
     popt, pcov = curve_fit(f=func, xdata=self.axes[0].centers, ydata=values)
--- a/src/hist/plot.py
+++ b/src/hist/plot.py
@@ -170,7 +170,12 @@ def plot_pull(
 
     # Computation and Fit
     values = self.values()
-    yerr = np.sqrt(self.variances())
+    variances = self.variances()
+    if variances is None:
+        raise RuntimeError(
+            "Cannot compute from a variance-less histogram, try a Weight storage"
+        )
+    yerr = np.sqrt(variances)
 
     # Compute fit values: using func as fit model
     popt, pcov = curve_fit(f=func, xdata=self.axes[0].centers, ydata=values)
--- a/src/hist/plot.py
+++ b/src/hist/plot.py
@@ -170,7 +170,7 @@ def plot_pull(
 
     # Computation and Fit
     values = self.values()
-    yerr = self.variances()
+    yerr = self.variances() ** 0.5
 
     # Compute fit values: using func as fit model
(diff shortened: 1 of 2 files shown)


src/hist/plot.py
```diff
@@ -170,7 +170,7 @@ def plot_pull(
 
     # Computation and Fit
     values = self.values()
-    yerr = self.variances()
+    yerr = self.variances() ** 0.5
```
matthewfeickert, line 173: As `variances` is a NumPy array 
```suggestion
    yerr = np.sqrt(self.variances())
```

```
variances(flow: bool = False) -> Union[numpy.ndarray, NoneType] method of hist.hist.Hist instance
    Returns the estimated variance of the accumulated values. The sum of squared
    weights for weighted histograms, the variance of samples for profiles, etc.
    For an unweighed histogram where kind == "COUNT", this should return the same
    as values if the histogram was not filled with weights, and None otherwise.
    If counts is equal to 1 or less, the variance in that cell is undefined if
    kind == "MEAN". This must be written <= 1, and not < 2; when this
    effective counts (weighed mean), then counts could be less than 2 but
    more than 1.
    
    If kind == "MEAN", the counts can be used to compute the error on the mean
    as sqrt(variances / counts), this works whether or not the entries are
    weighted if the weight variance was tracked by the implementation.
    
    Currently, this always returns - but in the future, it will return None
    if a weighted fill is made on a unweighed storage.
    
    :param flow: Enable flow bins. Not part of PlottableHistogram, but
    included for consistency with other methods and flexibility.
    
    :return: np.ndarray[np.float64]
```

can we use `np.sqrt` here instead?

```
$ ipython
Python 3.8.7 (default, Feb 13 2021, 01:38:22) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.21.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np

In [2]: import timeit

In [3]: x = np.arange(100)

In [4]: %timeit np.sqrt(x)
710 ns ± 7.58 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit x ** 0.5
2.7 µs ± 38.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

henryiii: That's disappointing. Yes, we can.

matthewfeickert: I guess it will hardly matter for small arrays though, and most histograms will have under 100 entries.

henryiii: Not that this is likely to be performance critical. I'm sure the plotting part is much slower. But we should follow good practice.

matthewfeickert: Sounds good to me. I hadn't actually done this test before today, but I had remembered seeing this post in 2019 https://twitter.com/HEPfeickert/status/1136730519226372096?s=20 about `math.pow(x, 2)` is more efficient than `x ** 2` and so did a quick check that my hunch that `np.sqrt` might also be faster than `x ** 0.5`.

PR #160, opened 2021-03-18
feat: adding profile for COUNT -> MEAN conversion

### LINKED ISSUE #156: [FEATURE] profile histograms
I personally find myself constructing profiles (`TH2::ProfileX()/TH2::ProfileY()`) during exploratory analysis. Is this something best left to the user, or would `hist` benefit from having the ability to do something like `h.profile(0).plot()`? To consider a specific case of 2 numerical axes,

```python
import numpy as np

def calculate_profile(self, axis=0):
    # matches the default error option for `TH2::ProfileX/Y()`
    values = self.values()
    variances = self.variances()
    centers = self.axes[axis].centers
    if axis == 0:
        num = np.matmul(values, centers)
        num_err = np.sqrt(np.matmul(variances, centers ** 2))
    else:
        num = np.matmul(values.T, centers)
        num_err = np.sqrt(np.matmul(variances.T, centers ** 2))
    den = np.sum(values.T, axis=axis)
    den_err = np.sqrt(np.sum(variances.T, axis=axis))
    new_values = num / den
    new_variances = ((num_err / den) ** 2 - (den_err * num / den ** 2.0) ** 2.0)
    return new_values, new_variances
```
(apologies for the convoluted numpy operations. I'm sure it can be simplified with some more thought)
and
```python
import hist

np.random.seed(42)
xy = np.random.multivariate_normal([0, 1], cov=[[1.,0.8],[0.8,1]], size=1000)
h = hist.Hist(hist.axis.Regular(50, -3, 3),hist.axis.Regular(50, -3, 3)).fill(*xy.T)

fig, ax = plt.subplots()
h.plot()

profile_axis = 0
values, variances = calculate_profile(h, axis=profile_axis)
ax.errorbar(h.axes[profile_axis].centers, values, yerr=variances**0.5, marker=".", color="r", alpha=0.5)
```
would yield
image

henryiii:
Closes #156, adds a `.profile` method. Based heavily on @aminnj's example.

TODO:
* [x] Needs tests
* This probably could be a WeightedMean, probably give a choice? Just got the basics working for now. (Can revisit later)
* [x] Should hide or handle warnings (true for some of our tests, too)

Followup:
* histoprint should handle showing kind=MEAN histograms (it doesn't like the NaN's, I think) CC @ast0815 (edit: done!)
* mplhep should show a better plot style for kind=MEAN, and should also ensure it handles kind=MEAN's requirements (we could also dispatch differently, but the best fix would be in mplhep) CC @andrzejnovak
* Boost-histogram should describe accumulators/storages a bit more in the docs, especially setting with a stack, including the correct order. CC Me

If kind=MEAN, you are only supposed to plot variances if `counts() > 1`, and values if `counts() >= 1`.

### DIFF
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -310,7 +310,7 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
         den_err = np.sqrt(np.sum(variances, axis=axis))
 
         new_values = num / den
-        new_variances = (num_err / den) ** 2 + (den_err * num / den ** 2) ** 2
+        new_variances = (num_err / den) ** 2 - (den_err * num / den ** 2) ** 2
 
         retval = self.__class__(*axes, storage=hist.storage.Mean())
         retval[...] = np.stack([count, new_values, new_variances], axis=-1)
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -282,6 +282,40 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
 
         return super().__setitem__(self._index_transform(index), value)
 
+    def profile(self: T, axis: Union[int, str]) -> T:
+        """
+        Returns a profile (Mean/WeightedMean) histogram from a normal histogram
+        with N-1 axes. The axis given is profiled over and removed from the
+        final histogram.
+        """
+
+        if self.kind != bh.Kind.COUNT:
+            raise TypeError("Profile requires a COUNT histogram")
+
+        axes = list(self.axes)
+        iaxis = axis if isinstance(axis, int) else self._name_to_index(axis)
+        axes.pop(iaxis)
+
+        values = self.values()
+        tmp_variances = self.variances()
+        variances = tmp_variances if tmp_variances is not None else values
+        centers = self.axes[iaxis].centers
+
+        count = np.sum(values, axis=iaxis)
+
+        num = np.tensordot(values, centers, ([iaxis], [0]))
+        num_err = np.sqrt(np.tensordot(variances, centers ** 2, ([iaxis], [0])))
+
+        den = np.sum(values, axis=axis)
+        den_err = np.sqrt(np.sum(variances, axis=axis))
+
+        new_values = num / den
+        new_variances = (num_err / den) ** 2 + (den_err * num / den ** 2) ** 2
+
+        retval = self.__class__(*axes, storage=hist.storage.Mean())
+        retval[...] = np.stack([count, new_values, new_variances], axis=-1)
+        return retval
+
     def density(self) -> np.ndarray:
         """
         Density NumPy array.
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -309,8 +309,9 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
         den = np.sum(values, axis=iaxis)
         den_err = np.sqrt(np.sum(variances, axis=iaxis))
 
-        new_values = num / den
-        new_variances = (num_err / den) ** 2 - (den_err * num / den ** 2) ** 2
+        with np.errstate(invalid="ignore"):
+            new_values = num / den
+            new_variances = (num_err / den) ** 2 - (den_err * num / den ** 2) ** 2
 
         retval = self.__class__(*axes, storage=hist.storage.Mean())
         retval[...] = np.stack([count, new_values, count * new_variances], axis=-1)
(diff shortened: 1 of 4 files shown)


src/hist/basehist.py
```diff
@@ -282,6 +282,40 @@ def __setitem__(  # type: ignore
...
+        count = np.sum(values, axis=iaxis)
+
+        num = np.tensordot(values, centers, ([iaxis], [0]))
+        num_err = np.sqrt(np.tensordot(variances, centers ** 2, ([iaxis], [0])))
+
+        den = np.sum(values, axis=axis)
+        den_err = np.sqrt(np.sum(variances, axis=axis))
+
+        new_values = num / den
+        new_variances = (num_err / den) ** 2 + (den_err * num / den ** 2) ** 2
```
aminnj, line 313: After checking consistency between ROOT and numpy, the error bars were not quite right. I updated https://github.com/scikit-hep/hist/issues/156 just now to fix the sign that I messed up, . This line should be
```python
new_variances = (num_err / den) ** 2 - (den_err * num / den ** 2) ** 2
```
(sign flipped in the middle)
Sorry...

aminnj:
After fixing the sign above, I get consistency between the function I have in https://github.com/scikit-hep/hist/issues/156 as well as ROOT (`TH2F::ProfileX()`), but the error bars from `h.profile(1).plot(label="hist")` look to be too small:
image.
I'm not sure what the issue could be. I verified that the calculations up to, and including, `new_variances` in
 https://github.com/scikit-hep/hist/pull/160/files/3a290a9f8fe2a7023686a74c80e46c8fd1d9929e#diff-dd1c19dfb377c8c01bb8892201933a2101eef8d01c705cb55ab3c93530cfec78R313-R317 
are correct, but the return histogram from `h.profile(1)` has variances that are too small.

aminnj:
Here's a (currently failing) test case.
```python
import hist
import numpy as np

xy = np.array([
    [-2, 1.5], [-2, -3.5], [-2, 1.5], # x = -2
    [0., -2.], [0., -2.], [0., 0.], [0., 2.], [0., 4.], # x = 0
    [2, 1.5], # x = +2
])
h = hist.Hist(hist.axis.Regular(5, -5, 5),hist.axis.Regular(5, -5, 5)).fill(*xy.T)

# Profile out the y-axis
hp = h.profile(1)

# Exclude edge bins since no values from above will fall into them
# When there are values there, ROOT does something funky in those bins,
# despite these bins not being in the axis that is profiled out, and
# despite there being no overflow... to be understood.
print("hist values:   ", hp.values()[1:-1])
print("hist variances:", hp.variances()[1:-1])
# hist values:    [-1.48029737e-16  4.00000000e-01  2.00000000e+00]
# hist variances: [0.88888889 0.2176            nan]


print("root values:   ", np.array([0.0, 0.4, 2.0]))
print("root variances:", np.array([2.66666667, 1.088, 0.0]))
# root values:    [0.  0.4 2. ]
# root variances: [2.66666667 1.088      0.        ]
```

The ROOT part is
```python
import ROOT
h = ROOT.TH2F("h","", 5,-5,5, 5,-5,5)
for x,y in xy:
    h.Fill(x,y)
hp = h.ProfileX()
for i in [2,3,4]:
    print(hp.GetBinContent(i), hp.GetBinError(i)**2)
```

src/hist/basehist.py
henryiii:
```suggestion
        new_variances = (num_err / den) ** 2 - (den_err * num / den ** 2) ** 2
```

henryiii:
I added a property-based test, which kept failing, and finally highlighted the mistake - my mistake, will fix. (I'll probably put the pbt in boost-histogram, rather than here).

src/hist/basehist.py
```diff
@@ -282,6 +282,40 @@ def __setitem__(  # type: ignore
...
+        num_err = np.sqrt(np.tensordot(variances, centers ** 2, ([iaxis], [0])))
+
+        den = np.sum(values, axis=axis)
+        den_err = np.sqrt(np.sum(variances, axis=axis))
+
+        new_values = num / den
+        new_variances = (num_err / den) ** 2 - (den_err * num / den ** 2) ** 2
+
+        retval = self.__class__(*axes, storage=hist.storage.Mean())
+        retval[...] = np.stack([count, new_values, new_variances], axis=-1)
```
henryiii, line 316:
```suggestion
        retval[...] = np.stack([count, new_values, count*new_variances], axis=-1)
```

henryiii:
Now that should be closer?

aminnj:
Indeed, errors look perfect now.
image

For the test case above, I get
```
hist variances: [2.66666667 1.088             nan]
root variances: [2.66666667 1.088      0.        ]
```
I guess the `0` vs `nan` is a convention. There's only 1 value from which to compute a std in that slice, so if nan suppresses the error bar when plotting, then I don't see a practical difference between the two.

ast0815:
I have just released v2.0.1 of Histoprint. It should now work with `nan` in the values. They are treated as `0`.

PR #161, opened 2021-03-20
feat: Add ratio plot support through .plot_ratio API

### LINKED ISSUE #148: [FEATURE] Add ratio plot support to API

matthewfeickert:
Resolves #148 

This PR adds support for ratio plots to the `.plot...` API by adding `.plot_ratio` and refactoring `.plot_pull` to use `.plot_ratio`.

This PR is still very rough, but I thought I'd open it up as there is a minimal working example now and then revise it heavily from there.

As the commit log is huge and messy I'll rebase and squash things into more reasonable commit sections to make review more digestible.

## Questions to come to consensus on
(Discussion can happen on Issue #148)

- Should the interval functions be moved to their own `stat_interval` module?

Answer: Yes, this is PR #176.

## TODO before requesting review

- [x] Finish refactoring `.plot_pull`
- [x] Add kwargs and kwargs filtering support
- [x] Fix typing errors
- [x] Add tests
- [x] All tests are passing
- [x] Docstrings added
- [x] README updated
- [x] Changelog updated.

## Suggested squash and merge message

```
* Add .plot_ratio to BastHist API
   - Adds ratio plot support for other Hists and callables
* Factor out plot_ratio and plot_pull to perform the final subplot plotting
   - Majority of logic is factored out of plot_pull into _plot_ratiolike
   - _plot_ratiolike then calls plot_ratio or plot_pull as needed
* Add plotting tests and test images for plot_ratio
* Update test image for plot_pull
* Update README and Changelog to reflect addition of plot_ratio

Co-authored-by: Henry Schreiner <henryschreineriii@gmail.com>
```


matthewfeickert:
Minimal toy working example before beginning revision work:

```python
# example.py
import matplotlib.pyplot as plt
import numpy as np

import hist

if __name__ == "__main__":
    np.random.seed(0)

    plt.figure(figsize=(10, 8))

    n_fill_1 = 5000
    n_fill_2 = n_fill_1 * 1.7

    hist_1 = hist.Hist(
        hist.axis.Regular(
            50, -5, 5, name="X", label="x [units]", underflow=False, overflow=False
        )
    ).fill(np.random.normal(size=int(n_fill_1)))
    hist_2 = hist.Hist(
        hist.axis.Regular(
            50, -5, 5, name="X", label="x [units]", underflow=False, overflow=False
        )
    ).fill(np.random.normal(size=int(n_fill_2)))

    plt.figure(figsize=(10, 8))
    ax, ratio_ax = hist_1.plot_ratio(hist_2)
    ax.get_figure().savefig("hist_ratio_plot.png")
```

hist_ratio_plot

Support for both callables and hists exists.

src/hist/plot.py
```diff
@@ -212,6 +220,309 @@ def plot2d_full(
...
+        if variances is None:
+            raise RuntimeError(
+                "Cannot compute from a variance-less histogram, try a Weight storage"
+            )
+        numerator_uncert = np.sqrt(variances)
+
+        # Infer model parameters using `other` as fit model
+        popt, pcov = _curve_fit_wrapper(
+            other, x_values, numerator, numerator_uncert, likelihood=likelihood
+        )
```
matthewfeickert, line 404: Question about types: If I run pre-commit I get 

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

src/hist/plot.py:403: error: Argument 1 to "_curve_fit_wrapper" has incompatible type "Union[BaseHist, Callable[[ndarray], ndarray], str]"; expected "Callable[..., Any]"
src/hist/plot.py:405: error: "BaseHist" not callable
src/hist/plot.py:405: error: "str" not callable
src/hist/plot.py:410: error: "BaseHist" not callable
src/hist/plot.py:410: error: "str" not callable
src/hist/plot.py:417: error: Item "str" of "Union[BaseHist, str]" has no attribute "values"
Found 6 errors in 1 file (checked 17 source files)
```

but I don't understand why I'm getting these error given the control flow (if this code is being executed then `other` is a callable and not a `hist.Basehist`). This need to get refactored anyway, but to get it to stop yelling at me Is it not possible to have type dependent control flow if you're using type checking in Python?

henryiii: Never use `type(other) in [str]`. Don't even use `type(other) in {str}`. Never directly compare types, always use `isinstance(other, str)`. Some flake8 checks will force this, I think.  This is why mypy can't type narrow on str as well, but not the only reason. It's perfectly valid to subclass str, but your version would break. For callable, it probably can't distinguish that callable will filter out BaseHist (or it might be fixed by the above fix). You can check `isinstance(other, BaseHist)`, then the narrowing should work.

src/hist/plot.py
```diff
@@ -212,16 +220,230 @@ def plot2d_full(
...
+        ratio_uncert = np.abs(clopper_pearson_interval(num, num + denom) - ratio)
+    return ratio_uncert
+
+
+# TODO: FIX TYPING
+def _fit_callable_to_hist(
+    model: Callable[[np.ndarray], np.ndarray],
+    _hist: hist.BaseHist,
+    likelihood: bool = False,
+) -> "Tuple[np.ndarray, np.ndarray, np.ndarray, Tuple[Tuple[np.float64, ...], np.ndarray]]":
```
matthewfeickert, line 325: I have confused myself on typing again as I don't understand why my tuple with a tuple of floats is wrong.

```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

src/hist/plot.py:351: error: Incompatible return value type (got "Tuple[ndarray, Any, Any, Tuple[Tuple[float, ...], Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]]]", expected "Tuple[ndarray, ndarray, ndarray, Tuple[Tuple[floating[_64Bit], ...], ndarray]]")
Found 1 error in 1 file (checked 17 source files)
```

henryiii:
```suggestion
) -> Tuple[np.ndarray, Any, Any, Tuple[Tuple[float, ...], "ArrayLike"]]:
```

Should be just fine. But better yet, fix the Any's and ArrayLike, they are too broad for a good return.

henryiii: Rebase or merge master in, and this should be better. I've swapped ArrayLike for np.ndarray, it always was an ndarray anyway.

src/hist/plot.py
```diff
@@ -210,16 +218,229 @@ def plot2d_full(
...
+) -> "Tuple[np.ndarray, np.ndarray, np.ndarray, Tuple[Tuple[float, ...], np.ndarray]]":
+    """
+    Fit a model, a callable function, to the histogram values.
+    """
+    variances = _hist.variances()
+    if variances is None:
+        raise RuntimeError(
+            "Cannot compute from a variance-less histogram, try a Weight storage"
+        )
+    hist_uncert = np.sqrt(variances)
```
henryiii, line 331:
```suggestion
    hist_uncert: np.ndarray = np.sqrt(variances)
```

Sadly needed, due to https://github.com/numpy/numpy/blob/c1ce397565398dcf22eda23f37d3c77ffe8f9b13/numpy/__init__.pyi#L2925 

You'll also need to predeclare `model_uncert: np.ndarray`. You can use cast or # type: ignore to "fix" the fact you are setting a typed value with Any.

matthewfeickert: Sorry, can you explain why this is needed in more detail? I believe you, but I don't understand given that things are passing `mypy` and I'm trying to better get the rules for typing.

matthewfeickert:
@henryiii This PR still isn't done, as I still have plenty to do on the checklist I added in the PR body, but after I rebase this off PR #168 if you have time to give general thoughts and feedback on better design of the refactor and clarity of control flow I would be very happy to make any and all suggested revisions while adding tests and docs.

src/hist/plot.py
henryiii: Well, if MyPy is happy, go with it. :) But in general:

```python
from typing import Any

def f(x: Any) -> int:
    return x
```
fails with:
```
tmp.py:4: error: Returning Any from function declared to return "int"
```

It's generally an error to return a more specific type from something that's less specific. However, if you manually set it in a single line, it's happy:

```python
from typing import Any

def f(x: Any) -> int:
    y: int = x
    return y
```
(I think the RHS here must be very simple to be treated like a cast.)
Or if you cast:
```python
import typing
from typing import Any

def f(x: Any) -> int:
    return typing.cast(int, x)
```
Not as fond of casting personally, as it gets in the way of runtime code more than the explicit assignment or a type ignore.

matthewfeickert: Thanks! I really appreciate you taking the time to explain. I'll change this when I'm out of meetings, but this helps me understand better!

src/hist/plot.py
```diff
@@ -210,19 +218,233 @@ def plot2d_full(
...
+
+
+def _plot_ratiolike(
     self: hist.BaseHist,
-    func: Union[Callable[[np.ndarray], np.ndarray], str],
+    other: Union[hist.BaseHist, Callable[[np.ndarray], np.ndarray]],
     likelihood: bool = False,
     *,
     ax_dict: "Optional[Dict[str, matplotlib.axes.Axes]]" = None,
+    view: str,
```
henryiii, line 440:
```suggestion
    view: Literal["ratio", "view"],
```

but, better yet, can't pull just call ratio, and then modify a bit if needed, rather than having them merged like this? Having factored out functions is better, and it seems like you were mostly there.

src/hist/plot.py
henryiii: If MyPy is currently happy, I wouldn't worry too much about it. This is the general idea, though.

src/hist/plot.py
henryiii: You'd have a plot_fit, a plot_ratio, then plot_pull would make two axes (or be given two axes) and then plot_fit on one, and plot_ratio on the other.

matthewfeickert: Oh, I see the CI is failing as `mypy` is complaining as [`typing.Literal` was added in Python 3.8](https://docs.python.org/3/library/typing.html#typing.Literal). I assume there's some way to get things to work with older Pythons, so I'll fix this.

As for having `plot_pull` call `plot_ratio` I'm a bit confused as how that would work given that `plot_pull` and `plot_ratio` are plotting different quantities in their subplots. I guess I could see something like `plot_pull` and `plot_ratio` both calling things like `plot_fit` (and some other function to just plot the two histograms for a ratio of two hists) to populate the main axis, and then both calling a revised `plot_ratiolike` that populates the subplot with the relevant type of plot (pull or ratio).

matthewfeickert:
(Copying from https://github.com/scikit-hep/hist/pull/161#discussion_r600634768 to make it easier for me to find later)

> but, better yet, can't pull just call ratio, and then modify a bit if needed, rather than having them merged like this? Having factored out functions is better, and it seems like you were mostly there. You'd have a `plot_fit`, a `plot_ratio`, then `plot_pull` would make two axes (or be given two axes) and then `plot_fit` on one, and `plot_ratio` on the other.

Oh, I see the CI is failing as `mypy` is complaining as [`typing.Literal` was added in Python 3.8](https://docs.python.org/3/library/typing.html#typing.Literal). I assume there's some way to get things to work with older Pythons, so I'll fix this.

As for having `plot_pull` call `plot_ratio` I'm a bit confused as how that would work given that `plot_pull` and `plot_ratio` are plotting different quantities in their subplots. I guess I could see something like `plot_pull` and `plot_ratio` both calling things like `plot_fit` (and some other function to just plot the two histograms for a ratio of two hists) to populate the main axis, and then both calling a revised `plot_ratiolike` that populates the subplot with the relevant type of plot (pull or ratio).

matthewfeickert:
> Having factored out functions is better, and it seems like you were mostly there.

After doing some more refactoring (perhaps too much?), this PR is getting pretty ugly looking in the review panels. If this is too much (this is definitely no longer atomic) I could make another PR that would do the refactoring of things first and then fix this PR to just do the ratio plot and pull plot work.

> You'd have a `plot_fit`, a `plot_ratio`, then `plot_pull` would make two axes (or be given two axes) and then `plot_fit` on one, and `plot_ratio` on the other.

As the pulls and the ratios have little overlap in what is actually being plotted (beyond that they are both in a subplot) the current setup has the `hist.BaseHist` `.plot_ratio` and `.plot_pull` functions call `_plot_ratiolike`. `_plot_ratiolike` accepts or creates the two axes, manages the calls to be able to fit the callable model to the hist if needed, and calls plotting for the main axis. This is all needed to make both ratio plots and pull plots, so this logic is used in both cases. The plotting of either the ratios or the pulls in the subplot is what `plot_ratio` or `plot_pull` manages, which don't really share any logic.

Thoughts? Am I missing something about the logic/design here?

henryiii:
The GitHub rich diff tools for images are quite nice, hadn't played with them before (there are three modes, for example). The image changes do not look very good; the band is darker, hiding the function plot, and the data points now plot over the legend.

src/hist/plot.py
```diff
@@ -6,12 +6,18 @@
 
 import hist
 
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
```
henryiii, line 12: Generally I like to bundle these conditional imports into a `typing.py` file in the package, but I can do that later.

src/hist/plot.py
```diff
@@ -210,159 +223,334 @@ def plot2d_full(
...
-
-        else:
-            func = _expr_to_lambda(func)
-
-    assert not isinstance(func, str)
-
-    parnames = list(inspect.signature(func).parameters)[1:]
+    interval_min = stats.beta.ppf((1 - coverage) / 2, num, denom - num + 1)
+    interval_max = stats.beta.ppf((1 + coverage) / 2, num + 1, denom - num)
+    interval = np.array([interval_min, interval_max])
```
henryiii, line 298: Is this actually a np.stack?

src/hist/plot.py
```diff
@@ -210,159 +223,334 @@ def plot2d_full(
...
-    if np.isfinite(pcov).all():
-        nsamples = 100
-        vopts = np.random.multivariate_normal(popt, pcov, nsamples)
-        sampled_ydata = np.vstack([func(xdata, *vopt).T for vopt in vopts])
-        yfiterr = np.nanstd(sampled_ydata, axis=0)
-    else:
-        yfiterr = np.zeros_like(yerr)
-
-    # Compute pulls: containing no INF values
+# TODO: Why is return type "Any"?
```
henryiii, line 304: Because ufuncts return Any? I need to fix a couple of Any return types before the next release anyway, I think.

src/hist/plot.py
```diff
@@ -210,159 +223,334 @@ def plot2d_full(
...
+    x_values = _hist.axes[0].centers
+    hist_values = _hist.values()
+
+    # gaussian with reasonable initial guesses for parameters
+    constant = float(hist_values.max())
+    mean = (hist_values * x_values).sum() / hist_values.sum()
+    sigma = (hist_values * np.square(x_values - mean)).sum() / hist_values.sum()
+
+    # TODO: Figure out why type must be ignored to pass mypy
+    # error: Returning Any from function declared to return "ndarray"
```
henryiii, line 371: Same issue, NumPy returns Any from ufuncts. Eventually this will be improved, I hope, but that's what we have for now.

src/hist/plot.py
```diff
@@ -376,19 +564,149 @@ def func(
...
+    r"""
+    Plot ratio like plots for BaseHist
+
+    ``fit_fmt`` can be a string such as ``r"{name} = {value:.3g} $\pm$ {error:.3g}"``
+    """
+
+    try:
+        from iminuit import Minuit  # noqa: F401
+        from scipy.optimize import curve_fit  # noqa: F401
+    except ImportError:
```
henryiii, line 601:
```suggestion
    except ModuleNotFoundError:
```

src/hist/plot.py
```diff
@@ -376,19 +564,149 @@ def func(
...
+        )
+        raise
+
+    if self.ndim != 1:
+        raise TypeError(
+            f"Only 1D-histogram supports ratio plot, try projecting {self.__class__.__name__} to 1D"
+        )
+    if isinstance(other, hist.hist.Hist) and other.ndim != 1:
+        raise TypeError(
+            f"Only 1D-histogram supports ratio plot, try projecting {other.__class__.__name__} to 1D"
```
henryiii, line 614:
```suggestion
            f"Only 1D-histogram supports ratio plot, try projecting other={other.__class__.__name__} to 1D"
```

src/hist/plot.py
```diff
@@ -376,19 +564,149 @@ def func(
...
+            subplot_ax = plot_ratio(
+                self, ratios, ratio_uncert, ax=subplot_ax, **rp_kwargs
+            )
+
+        elif view == "pull":
+            pulls = (hist_values - compare_values) / hist_values_uncert
 
-    plt.xlim(left_edge, right_edge)
+            pulls[np.isnan(pulls)] = 0
+            pulls[np.isinf(pulls)] = 0
```
henryiii, line 700:
```suggestion
            pulls[np.isnan(pulls) | np.isinf(pulls)] = 0
```

henryiii (approved):
Looks good, I had a few very minor comments. After this goes in, I might try to do a tiny bit of cleanup. (typing.py, fix an Any or two)

src/hist/plot.py
henryiii: Oh, I take that back, there is one already. Please move this to src/hist/typing.py and import from there.

matthewfeickert:
> The GitHub rich diff tools for images are quite nice, hadn't played with them before (there are three modes, for example). The image changes do not look very good; the band is darker, hiding the function plot, and the data points now plot over the legend.

I had never used them before either but that's actually _really nice_! :sparkles: I'll have to use this more. Wrt the images, I'll have to figure out why things looks so terrible color-wise as I get different results when I run the same code as a standalone script, so something is going wrong with the color wheel in the tests.

src/hist/plot.py
matthewfeickert: Yup. This was [a copypaste from `coffea`](https://github.com/CoffeaTeam/coffea/blob/12228d5564963234434d5c881efa761e088c298a/coffea/hist/plot.py#L75), but a good catch and example of some things I might want to clean up in those.

tests/test_plot.py
```diff
@@ -620,6 +620,12 @@ def pdf(x, a=1 / np.sqrt(2 * np.pi), x0=0, sigma=1, offset=0):
 
     fig = plt.figure()
 
-    assert h.plot_pull(pdf, fit_fmt=r"{name} = {value:.3g} $\pm$ {error:.3g}")
+    assert h.plot_pull(
+        pdf,
+        eb_color="black",
+        fp_color="blue",
+        ub_color="lightblue",
```
matthewfeickert, line 627: The default  `matplotlib` proc cycle colors are quite ugly for a plot like this, so choosing these for now so that the different components are actually visible.

andrzejnovak:
> The GitHub rich diff tools for images are quite nice,

Where does one access this?

src/hist/plot.py
```diff
@@ -210,159 +220,334 @@ def plot2d_full(
...
+        coverage : float, optional
+            Central coverage interval, defaults to 68%
+    Calculates the so-called 'Garwood' interval,
+    c.f. https://www.ine.pt/revstat/pdf/rs120203.pdf or
+    http://ms.mcmaster.ca/peter/s743/poissonalpha.html
+    For weighted data, this approximates the observed count by ``values**2/variances``, which
+    effectively scales the unweighted poisson interval by the average weight.
+    This may not be the optimal solution: see https://arxiv.org/pdf/1309.1287.pdf for a proper treatment.
+    When a bin is zero, the scale of the nearest nonzero bin is substituted to scale the nominal upper bound.
+    If all bins zero, a warning is generated and interval is set to ``values``.
```
nsmith-, line 246: This warning when everything is zero has been criticized by coffea users, as its a bit unwelcome and frequent (since empty histograms happen often).

matthewfeickert: @nsmith- yeah, that's why I later down have it called under a `with np.errstate(divide="ignore"):` context.

matthewfeickert:
> > The GitHub rich diff tools for images are quite nice,
> 
> Where does one access this?

@andrzejnovak It isn't well documented but for image files in the review mode there is a view button (turned off by default) that you can toggle on as seen in this example (with it on).

example_rich_diff

It allows you to do quite a lot as well for nice image comparison: https://docs.github.com/en/github/managing-files-in-a-repository/rendering-and-diffing-images

matthewfeickert:
@henryiii This is probably finally ready for review. It is huge and I feel bad about that :grimacing: so if you would prefer for me to split out parts of it into more atomic PRs (like the frequentist interval support) I'm more than happy to do so as (I know reviewing PRs that have diffs larger than 500 lines is unpleasant and not very productive).

If it would also help to squash it down into a few sections of work so that there aren't a huge number of commits I'm happy to do that as well.

I've also added as suggested squash and merge message up in the PR body, but as I'm not sure how you like the messages to be structured I'm happy to revise that to be more verbose or much more concise.

src/hist/intervals.py
```diff
@@ -0,0 +1,107 @@
+from typing import Any, Optional
+
+import numpy as np
+from scipy import stats
+
+__all__ = ("poisson_interval", "clopper_pearson_interval", "ratio_uncertainty")
+
+
+# Code originally contributed to coffea by Nicholas Smith
+# Copyright (c) 2018, Fermilab
```
matthewfeickert, line 10: @nsmith- If you want to review the intervals module that'd be cool too. :+1:

src/hist/intervals.py
```diff
@@ -0,0 +1,107 @@
...
+
+def ratio_uncertainty(
+    num: np.ndarray, denom: np.ndarray, uncert_type: str = "poisson"
+) -> Any:
+    # Note: As return is a numpy ufuncs the type is "Any"
+    with np.errstate(divide="ignore"):
+        ratio = num / denom
+    if uncert_type == "poisson":
+        ratio_uncert = np.abs(poisson_interval(ratio, num / np.square(denom)) - ratio)
+    elif uncert_type == "poisson-ratio":
```
henryiii, line 104: Can we error if an unknown `uncert_type` is passed? Also, why not `uncertainty_type`? If someone does type it out, why save 5 chars to make it harder to read?

src/hist/intervals.py
```diff
@@ -0,0 +1,107 @@
...
+    interval_min = stats.beta.ppf((1 - coverage) / 2, num, denom - num + 1)
+    interval_max = stats.beta.ppf((1 + coverage) / 2, num + 1, denom - num)
+    interval = np.stack((interval_min, interval_max))
+    interval[:, num == 0.0] = 0.0
+    interval[1, num == denom] = 1.0
+    return interval
+
+
+def ratio_uncertainty(
+    num: np.ndarray, denom: np.ndarray, uncert_type: str = "poisson"
```
henryiii, line 97:
```suggestion
    num: np.ndarray, denom: np.ndarray, uncert_type: Literal["poisson", "poisson-ratio"] = "poisson"
```

henryiii: (Or could be an enum)

src/hist/intervals.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    if coverage is None:
+        coverage = stats.norm.cdf(1) - stats.norm.cdf(-1)
+    scale = np.empty_like(values)
+    scale[values != 0] = variances[values != 0] / values[values != 0]
+    if np.sum(values == 0) > 0:
+        missing = np.where(values == 0)
+        available = np.nonzero(values)
+        if len(available[0]) == 0:
+            raise RuntimeWarning(
```
henryiii, line 47:
```suggestion
            raise RuntimeError(
```

or a better error. Or use the `warnings.warn` module. You can't raise a Warning, that breaks the control flow, so it's not a "Warning".

src/hist/plot.py
```diff
@@ -210,185 +220,386 @@ def plot2d_full(
...
+) -> Callable[[np.ndarray, float, float, float], np.ndarray]:
+    x_values = _hist.axes[0].centers
+    hist_values = _hist.values()
+
+    # gaussian with reasonable initial guesses for parameters
+    constant = float(hist_values.max())
+    mean = (hist_values * x_values).sum() / hist_values.sum()
+    sigma = (hist_values * np.square(x_values - mean)).sum() / hist_values.sum()
+
+    def gauss(
```
henryiii, line 234: This can/should be pulled out; it does not need to be a closure. Also the return is np.ndarray, since x is an np.ndarray.

src/hist/plot.py
```diff
@@ -210,185 +220,386 @@ def plot2d_full(
...
+    # TODO: Probably set a better default color than the fit line color
+    ub_kwargs.setdefault("color", line.get_color())
+    ax.fill_between(
+        x_values,
+        model_values - model_uncert,
+        model_values + model_uncert,
+        **ub_kwargs,
+    )
+
+    return ax
```
henryiii, line 315: Generally, I'd return the Artists, not the Axes. The caller already has the axes, and you can get the axes from the Artists, too.

src/hist/plot.py
```diff
@@ -210,185 +220,386 @@ def plot2d_full(
...
+
+
+def plot_ratio(
+    _hist: hist.BaseHist,
+    ratio: np.ndarray,
+    ratio_uncert: np.ndarray,
+    ax: matplotlib.axes.Axes,
+    **kwargs: Any,
+) -> matplotlib.axes.Axes:
```
henryiii, line 324: Same, always return Artists. We generally bundle multiple (different) artists into a NamedTuple.

henryiii (commented):
A few small comments, then it looks good!

matthewfeickert:
Awesome. Thanks for the fast and clear review @henryiii! :+1: I'll get on this later today. I'm also going to factor out the `intervals` into a separate PR that would go in first to make this one be focused on plotting, unless you would really prefer to reduce the number of PRs you have to look at.

src/hist/intervals.py
nsmith-: lmk if you want it here or in a new pr

matthewfeickert: @nsmith- if you've got feedback now feel free to unload it here and I can migrate any discussion points to the new PR.

nsmith-: I haven't looked at it, seems easier to do it on the new one if its made, or here if not.

matthewfeickert: SGTM.

eduardo-rodrigues: Allow me to chime in - is this copyright statement not going to "blow you in the face" down the line? First off, why would it not simply be copyright Nicholas Smith and why Fermilab, really? Would it not be better/safer to simply acknowledge the code from ... It's not been a week since I heard all kinds of worries related to copyrights, so ... And you're sure that this is OK with respect to BSD-3? Again, better play it safe and only complicate if no other reason, and for a real benefit. My 2 cents. Sorry for the interjection.

nsmith-: coffea is BSD-3 as well (which is allowed for fnal stuff), so the code can be lifted with attribution and inclusion of the license. What's not clear is how to attribute and include the license. @lgray may have ideas

src/hist/plot.py
andrzejnovak: This is probably not the right place, but I wonder if we should find a middle ground, maybe return the Axes in the named tuple as well. While it's easier to get axes from artists than the other way around, it makes for funny code and you want the axes more often than the artists.

src/hist/intervals.py
eduardo-rodrigues: Ah, that's nicer / good to know. I'm not sure either. But scikit-learn has plenty of code authored by many, explicitly mentioned in the files, and that does not cause problems for the license itself, which is a single one for "the developers". I would have to dive a bit more to see anything subtle in there.

lgray: At a minimum you just need to include a URL for the license in coffea so you prove the code base is under it.
I'd just take the URL of the license in the coffea github repository and stick it here, should be good enough and then no possibility of blowing up in your face.

lgray: As far as specificity of inclusion I think it's nominally to the entity in the license in the original repo. So "Fermilab" is the copyright holder.  You can probably include names of authors for fun.

henryiii: I believe if you have the permissions of all the original authors, they can relicense it here, simliar to if they had just directly contributed it.

src/hist/plot.py
henryiii: I think it's best to match matplotlib, and matplotlib always and only returns Artists. Most of the time, if you care about the axes, you already have them.

src/hist/intervals.py
matthewfeickert: Yeah, the BSD3 nature of the coffea, specifically the 

> * Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
  
  part is what prompted this (and you're right @lgray, the inclusion of Nick's name was just to give him a hat tip).
  
> I believe if you have the permissions of all the original authors, they can relicense it here, simliar to if they had just directly contributed it.

Ah this is very good to know! As I'm going to factor this out in the service of more atomic PRs let me ping all of you again on this on the new PR in a bit (unless @lgray and @nsmith- you agree with Henry here and would prefer to not get further pings on this).

matthewfeickert: This is now PR #176

src/hist/plot.py
```diff
@@ -210,185 +220,386 @@ def plot2d_full(
...
+            downRect_startpoint, patch_width, patch_height, **pp_kwargs
+        )
+        ax.add_patch(downRect)
+
+    ax.set_xlim(left_edge, right_edge)
+
+    ax.set_xlabel(_hist.axes[0].label)
+    ax.set_ylabel("Pull")
+
+    return ax
```
matthewfeickert, line 443: Note to self: This should also be returning the axis artists like the above requests.

henryiii:
@all-contributors please add @matthewfeickert for code

src/hist/plot.py
matthewfeickert: In general, is there a more programmatic way to get the list (or I guess from below a NamedTuple) of Artists beyond manually bookkeeping? I assume that `ax.get_children()` is a bit too broad as I'm assuming you don't(?) want the `axis.XAxis` and `axis.YAxis` in there as well.

src/hist/plot.py
matthewfeickert: > This can/should be pulled out; it does not need to be a closure

Maybe I'm missing the obvious but then should we get rid of this `_construct_gaussian_callable` function and just have these lines lower down in `_plot_ratiolike`? I guess I don't see how to avoid a closure as we want all model parameters of `gauss` to be set now and then get `x` later as `xdata` in `_curve_fit_wrapper`.

> the return is np.ndarray, since x is an np.ndarray.

This is what I thought as well and then was confused by `mypy`'s response

```
src/hist/plot.py:245: error: Returning Any from function declared to return "ndarray"
```

:?

src/hist/plot.py
matthewfeickert: oh. `plot2d_full` shows how: Just name the returns and of the plot calls and return them at the end. :+1:

henryiii: We are basically matching what matplotlib does. You can't get the exact Artists associated with a plotting function call except by the return. There could already be artists on the axis.

We are enhancing it slightly via named tuple, but same spirit.

matthewfeickert: This was actually pretty educational in understanding the `matplotlib` API design better, or rather actually paying attenditon to what the API says it is doing. :grimacing: 

I still need to clean these up by using `namedtuple`, but this is hopefully slightly better shape now.

src/hist/plot.py
henryiii: That's because a UFunc returns Any, because following though the possible return types is hard. But you know you have an ndarray. You can force it with `retval: np.ndarray = <ufunc>; return retval`, or a `# type: ignore`.

henryiii: I might have missed that it really was a closure, that's fine if it is then.

henryiii: Ahh, it's a closure due to the default arguments.

henryiii:
This is probably the last thing going into 2.3, which will probably be the last Python 3.6 release.

matthewfeickert:
> This is probably the last thing going into 2.3, which will probably be the last Python 3.6 release.

:+1: I'm in meeting hell today, but I should hopefully be done by 17:00 Illinois time and then I can finish this.

src/hist/plot.py
matthewfeickert: I added a note and forced the return type like you suggested. :+1:

matthewfeickert:
@henryiii my lack of experience with `namedtuple` is showing here, but can you give me a TLDR of what using `namedtuple` like you do here in `mplhep`

https://github.com/scikit-hep/mplhep/blob/4f408d6e6bf8625ddaac912c7220e582f7c4d241/src/mplhep/plot.py#L24-L30

```python
StairsArtists = namedtuple("StairsArtists", "stairs errorbar legend_artist")
ErrorBarArtists = namedtuple("ErrorBarArtists", "errorbar")
ColormeshArtists = namedtuple("ColormeshArtists", "pcolormesh cbar")


Hist1DArtists = Union[StairsArtists, ErrorBarArtists]
Hist2DArtists = ColormeshArtists
```

 gives you compared to creating types from unions of types like you do in `hist` here (in terms of making it easier to give return types that are just `Tuple[Any, ...]`)?

https://github.com/scikit-hep/hist/blob/958037ffe46d4d2f9c04200a4540176712817c65/src/hist/basehist.py#L38-L42

henryiii:
```python
tuple[int | str, ...] # Tuple with int or str, of any length and combination
tuple[int, str] # tuple with the first element an int, second a str, always length 2
IntStr = namedtuple("IntStr", "a", "b")  # tuple length 2, with named access too
class IntStr(NamedTuple):  # Tuple length 2 with int, str AND named access
    a: int
    b: str
```
 The first example you showed was basically the third line (should be changed to a typed NamedTuple), while the second example is from the first line = just a tuple with a Union of possible items of any length.

matthewfeickert:
> ```python
> tuple[int | str, ...] # Tuple with int or str, of any length and combination
> tuple[int, str] # tuple with the first element an int, second a str, always length 2
> IntStr = namedtuple("IntStr", "a", "b")  # tuple length 2, with named access too
> class IntStr(NamedTuple):  # Tuple length 2 with int, str AND named access
>     a: int
>     b: str
> ```
> 
> The first example you showed was basically the third line (should be changed to a typed NamedTuple), while the second example is from the first line = just a tuple with a Union of possible items of any length.

Okay, I think this explains my confusion as it seemed like we were loosing the type information when going to `namedtuple`. So ideally to take advantage of the `matplotlib` types that I know that I need I should make classes from `typing.NamedTuple`. :+1:

src/hist/plot.py
```diff
@@ -214,185 +263,402 @@ def plot2d_full(
...
+            height=bar_height,
+            width=bar_width,
+            bottom=bar_bottom,
+            fill=False,
+            linewidth=0,
+            edgecolor="gray",
+            hatch=3 * "/",
+            zorder=bar_zorder,
+        )
+        axis_artists = RatioBarArtists(central_value_artist, _ratio_points, bar_artists)
```
matthewfeickert, line 423: @henryiii I'm getting errors (here and few other places for similar reasons) that 

```
src/hist/plot.py:423: error: Incompatible types in assignment (expression has type "RatioBarArtists", variable has type "RatioErrorbarArtists")
```

because in the control flow above for the other branch it is set as

```python
axis_artists = RatioErrorbarArtists(central_value_artist, errorbar_artists)
```

If I try and type both explicitly

```python
        axis_artists: RatioErrorbarArtists = RatioErrorbarArtists(
            central_value_artist, errorbar_artists
        )
        # ....
        axis_artists: RatioBarArtists = RatioBarArtists(
            central_value_artist, _ratio_points, bar_artists
        )
```

then I get 

```
src/hist/plot.py:425: error: Name 'axis_artists' already defined on line 398
```

Am I missing the obvious with how to deal with the control flow here? Or does this require type checking to be skipped at these lines?

henryiii:
```python
axis_artists: Union[RatioBarArtists, RatioErrorbarArtists]

axis_artists = RatioErrorbarArtists(
    central_value_artist, errorbar_artists
)
# ....
axis_artists = RatioBarArtists(
    central_value_artist, _ratio_points, bar_artists
)
```

matthewfeickert: Thanks.

matthewfeickert:
I think at this point it is safe to post an example of what usage looks like now:

```python
import matplotlib.pyplot as plt
import numpy as np

import hist


def pdf(x, a=1 / np.sqrt(2 * np.pi), x0=0, sigma=1, offset=0):
    return a * np.exp(-((x - x0) ** 2) / (2 * sigma ** 2)) + offset


if __name__ == "__main__":
    np.random.seed(0)

    hist_1 = hist.Hist(
        hist.axis.Regular(
            50, -5, 5, name="X", label="x [units]", underflow=False, overflow=False
        )
    ).fill(np.random.normal(size=1000))
    hist_2 = hist.Hist(
        hist.axis.Regular(
            50, -5, 5, name="X", label="x [units]", underflow=False, overflow=False
        )
    ).fill(np.random.normal(size=1700))

    artists = hist_1.plot_pull(
        "normal",
        eb_ecolor="steelblue",
        eb_mfc="steelblue",
        eb_mec="steelblue",
        eb_fmt="o",
        eb_ms=6,
        eb_capsize=1,
        eb_capthick=2,
        eb_alpha=0.8,
        fp_c="hotpink",
        fp_ls="-",
        fp_lw=2,
        fp_alpha=0.8,
        bar_fc="royalblue",
        pp_num=3,
        pp_fc="royalblue",
        pp_alpha=0.618,
        pp_ec=None,
        ub_alpha=0.2,
    )
    fig.savefig("pull_plot.png")

    artists = hist_1.plot_ratio(
        hist_2,
        rp_ylabel=r"Ratio",
        rp_num_label="hist1",
        rp_denom_label="hist2",
        rp_uncert_draw_type="bar",
    )
    fig.savefig("ratio_plot_hist.png")

    artists = hist_1.plot_ratio(
        pdf,
        rp_ylabel=r"Ratio",
        rp_num_label="hist1",
        rp_denom_label="hist2",
    )
    fig.savefig("ratio_plot_callable.png")

```

**pull_plot.png**

pull_plot

**ratio_plot_hist.png**

ratio_plot_hist

**ratio_plot_callable.png**

ratio_plot_callable

matthewfeickert:
(I had forgotten that I still had a TODO that was https://github.com/scikit-hep/hist/pull/161/commits/6b9e3efd66ada2178c905895f0e903f8fa73428c :grimacing: but now I've git grepped the diff and it is TODO free, so should _actually_ be ready for review now)

eduardo-rodrigues:
One little question from me, @matthewfeickert: an argument "gauss" was introduced recently for the most obvious/common use case of the pull plot, so that the user does not have to define that `pdf` function. Could you not do the same fo the ratio plot? That's not the case in the example above.

andrzejnovak:
I could be in minority again, but any chance the kwargs passed to the artist could be cleaned a little? Maybe instead of a large number of custom kwargs, pass dicts to each of the primary artists? Especially since you return the artists tuple, it easier to fix anything in post if especially needed, so I think the primary call should be a bit simpler.

henryiii:
I wasn't all that fond of the evaluatable string argument; I would prefer "normal" Python be used, but we should be consistent. I'd be okay to provide a "gauss" function somewhere (hist.gauss or better yet hist.functions.gauss, etc), then remove the strings option entirely, or to only allow "gauss/gaus" and not arbitrary evaulatable strings. But I might be in the minority there too. Let's be consistent, though.

@andrzejnovak That's the current design, let's not change that _right_ now. Maybe open a discussion? The problem is that these tend to get added directly in the function call, so passing a bunch of dicts is potentially more difficult than passing prefixed keywords (for the record, I think I'm on your side, though). I think you can find examples of both types of API in matplotlib. We could add `something_kw` and combine them, perhaps, etc.

src/hist/basehist.py
```diff
@@ -401,20 +403,45 @@ def plot2d_full(
 
         return hist.plot.plot2d_full(self, ax_dict=ax_dict, **kwargs)
 
+    def plot_ratio(
+        self,
+        other: Callable[[np.ndarray], np.ndarray],
+        *,
+        ax_dict: "Optional[Dict[str, matplotlib.axes.Axes]]" = None,
+        **kwargs: Any,
+    ) -> "Tuple[MainAxisArtists, RatiolikeArtists]":
```
henryiii, line 412: Totally not on you, but I'm pretty sure none of these quotes are needed. They should all go away anyway when we go to 3.7+.

src/hist/plot.py
```diff
@@ -20,7 +34,42 @@
...
+
+
+__all__ = (
+    "histplot",
+    "hist2dplot",
+    "plot2d_full",
+    "plot_ratio",
+    "plot_pull",
+    "plot_pie",
+)
```
henryiii, line 72: Minor, but I"d generally place `__all__` as close to the top as reasonably allowed (just below the imports), since it's handy for users looking in a file to see what is publicly available. (I can change this later, so no need to change)

henryiii:
> I would prefer "normal" Python be used, but we should be consistent.

Further on this, how hard would it be to use "formulate" (via an example, not a dependency!) to write a string here? If you could do `hist.stuff(formalate.from_numexpr("..."))`, then you could also do `hist.stuff(formalate.from_root("..."))`, etc. :) (I'm guessing formulate is or should be callable to do this? @mayou36 are you the formulate maintainer?)

src/hist/plot.py
matthewfeickert: I can get this now. :)

matthewfeickert: Okay moved them.

src/hist/basehist.py
matthewfeickert: Hm. With the following diff

```diff
diff --git a/src/hist/basehist.py b/src/hist/basehist.py
index 45c1665..21021d5 100644
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -409,7 +409,7 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
         *,
         ax_dict: "Optional[Dict[str, matplotlib.axes.Axes]]" = None,
         **kwargs: Any,
-    ) -> "Tuple[MainAxisArtists, RatiolikeArtists]":
+    ) -> Tuple[MainAxisArtists, RatiolikeArtists]:
         """
         ``plot_ratio`` method for ``BaseHist`` object.
```

my posted example fails with

```pytb
$ python example.py 
Traceback (most recent call last):
  File "example.py", line 4, in <module>
    import hist
  File "/home/feickert/Code/GitHub/Scikit-HEP/hist/src/hist/__init__.py", line 11, in <module>
    from .basehist import BaseHist
  File "/home/feickert/Code/GitHub/Scikit-HEP/hist/src/hist/basehist.py", line 62, in <module>
    class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
  File "/home/feickert/Code/GitHub/Scikit-HEP/hist/src/hist/basehist.py", line 412, in BaseHist
    ) -> Tuple[MainAxisArtists, RatiolikeArtists]:
NameError: name 'MainAxisArtists' is not defined
```

matthewfeickert:
>  an argument "gauss" was introduced recently for the most obvious/common use case of the pull plot, so that the user does not have to define that `pdf` function. Could you not do the same fo the ratio plot? That's not the case in the example above.

@eduardo-rodrigues This is already supported. `hist_1.plot_ratio("normal")` will work (or `"gauss"` or `"gaus"`), but I just didn't show it as such in the example as `plot_ratio` and `plot_pull` now share all the logic internals and only differ in what is plotted in the subplot &mdash; the main axis plot logic is the same for both.

> That's the current design, let's not change that right now. Maybe open a discussion

@andrzejnovak I would second @henryiii's suggestion here.

src/hist/basehist.py
henryiii: Ahh, it's because of the TYPE_CHECKING line above. Not sure if that's required (except for ellipsis, that's required), but that's why.

matthewfeickert:
@henryiii Just ping me if you think there's anything left that needs to be done here (I know this isn't the nicest diff to review given its size), but from my side I think this is done. :+1:

src/hist/basehist.py
```diff
@@ -401,20 +403,45 @@ def plot2d_full(
...
+
+        import hist.plot
+
+        return hist.plot._plot_ratiolike(
+            self, other, ax_dict=ax_dict, view="ratio", **kwargs
+        )
+
     def plot_pull(
         self,
         func: Callable[[np.ndarray], np.ndarray],
```
henryiii, line 428:
```suggestion
        func: Union[str, Callable[[np.ndarray], np.ndarray]],
```

?

src/hist/basehist.py
matthewfeickert: Yeah, I just assumed as there was

```python
    from mplhep.plot import Hist1DArtists, Hist2DArtists
```

under it as well that I should keep all types from plotting in there as well. I can move them out if that would be better.

src/hist/basehist.py
```diff
@@ -401,20 +403,45 @@ def plot2d_full(
 
         return hist.plot.plot2d_full(self, ax_dict=ax_dict, **kwargs)
 
+    def plot_ratio(
+        self,
+        other: Callable[[np.ndarray], np.ndarray],
```
henryiii, line 408:
```suggestion
        other: Union[str, Callable[[np.ndarray], np.ndarray]],
```

src/hist/basehist.py
matthewfeickert: Ah yeah, I guess that got missed. Strange that it didn't complain. :?

src/hist/basehist.py
henryiii: We can address later, I'm guessing this might be due to an older version missing those? Soon 3.4+ will be required, I expect, due to mplhep.

src/hist/plot.py
```diff
@@ -214,185 +262,408 @@ def plot2d_full(
...
+    ax.set_xlabel(_hist.axes[0].label)
+    ax.set_ylabel("Pull")
+
+    return PullArtists(bar_artists, patch_artists)
+
+
+def _plot_ratiolike(
     self: hist.BaseHist,
-    func: Union[Callable[[np.ndarray], np.ndarray], str],
+    other: Union[hist.BaseHist, Callable[[np.ndarray], np.ndarray]],
```
henryiii, line 508:
```suggestion
    other: Union[hist.BaseHist, Callable[[np.ndarray], np.ndarray], str],
```

henryiii:
If we do support strings, then it should be part of the types. We might want to revisit the idea of supporting strings before the final release, though.

src/hist/basehist.py
matthewfeickert: Addition of `str` is causing `mypy` to explode everywhere, so this is going to take some debugging. :/ I'll request another review once done, but this won't be fast.

henryiii:
You can always remove a type; mypy won't complain unless you have an example using that removed type. (We don't check tests, and I don't think there's a test with this). MyPy checks that what you say is valid, not that you cover all possibilities.

The explosion is probably either from something down the line, or from the type not being narrowed after it is removed. To manually narrow, use `assert`. Example, this will fail:

```python
def f(x: A | B) -> B:
    if isinstance(x, A):
        x = B(x)
    return x
```
MyPy can't tell that X is now only A. But you can, so you can manually narrow:

```python
def f(x: A | B) -> B:
    if isinstance(x, A):
        x = B(x)
    assert not isinstance(x, A)
    return x
```

Now MyPy is happy.

andrzejnovak:
> @andrzejnovak I would second @henryiii's suggestion here.

not that I want to burden an already large PR, happy to take this to a discussion, but if structural changes are being done that's the right time to make API changes, not two releases later

henryiii:
Merging this PR does not create a release, so we don't have to have mega-PRs. :)

At this point, I think we are discussing an addition to the API, removing the old way of doing things would possibly happen later (3.0).

henryiii:
Having API additions to improve compatibility with an eventual 3.0 should be ideally done before dropping Python 3.6, so 3.6 users can use the "modern" api and be compatible with 3.0, so yes, having the discussion now is good.

src/hist/plot.py
```diff
@@ -214,185 +262,408 @@ def plot2d_full(
...
+) -> RatioArtists:
+    """
+    Plot a ratio plot on the given axes
+    """
+    x_values = _hist.axes[0].centers
+    left_edge = _hist.axes.edges[0][0]
+    right_edge = _hist.axes.edges[-1][-1]
+
+    # Set 0 and inf to nan to hide during plotting
+    ratio[ratio == 0] = np.nan
```
henryiii, line 380: > This is already supported. `hist_1.plot_ratio("normal")` will work

Are you sure? I think MyPy is right here. And in general, "will work" should be a synonym for "is tested". :P (though, I'd also trust mypy checking here, that's about 70% as effective as a test for something like this)

src/hist/plot.py
```diff
@@ -214,185 +262,408 @@ def plot2d_full(
...
+        sampled_ydata = np.vstack([model(xdata, *vopt).T for vopt in vopts])
+        model_uncert = np.nanstd(sampled_ydata, axis=0)
+    else:
+        model_uncert = np.zeros_like(hist_uncert)
+
+    return model_values, model_uncert, hist_uncert, (popt, pcov)
+
+
+def _plot_fit_result(
+    _hist: hist.BaseHist,
```
henryiii, line 326: By the way, if you want a "position only" variable before Python 3.8, you should use `__hist` - mypy literally will consider this a position only variable; it's a convention used in typeshed for the standard library. Though in this case, `histogram` would probably be fine.

henryiii:
FYI, this is the changes I tried; no fixes but added strings / callables and removed quotes in plot.py (since matplotlib is available there) (in .patch form if you want to apply them):

```patch
diff --git a/src/hist/basehist.py b/src/hist/basehist.py
index 45c1665..cf281a6 100644
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -405,7 +405,7 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
 
     def plot_ratio(
         self,
-        other: Callable[[np.ndarray], np.ndarray],
+        other: Union[np.ndarray, str, Callable[[np.ndarray], np.ndarray]],
         *,
         ax_dict: "Optional[Dict[str, matplotlib.axes.Axes]]" = None,
         **kwargs: Any,
@@ -425,7 +425,7 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
 
     def plot_pull(
         self,
-        func: Callable[[np.ndarray], np.ndarray],
+        func: Union[str, Callable[[np.ndarray], np.ndarray]],
         *,
         ax_dict: "Optional[Dict[str, matplotlib.axes.Axes]]" = None,
         **kwargs: Any,
diff --git a/src/hist/plot.py b/src/hist/plot.py
index ed3d7b1..c861f37 100644
--- a/src/hist/plot.py
+++ b/src/hist/plot.py
@@ -182,9 +182,9 @@ def _curve_fit_wrapper(
 def plot2d_full(
     self: hist.BaseHist,
     *,
-    ax_dict: "Optional[Dict[str, matplotlib.axes.Axes]]" = None,
+    ax_dict: Optional[Dict[str, matplotlib.axes.Axes]] = None,
     **kwargs: Any,
-) -> "Tuple[Hist2DArtists, Hist1DArtists, Hist1DArtists]":
+) -> Tuple[Hist2DArtists, Hist1DArtists, Hist1DArtists]:
     """
     Plot2d_full method for BaseHist object.
 
@@ -364,7 +364,7 @@ def _plot_fit_result(
 
 def plot_ratio(
     _hist: hist.BaseHist,
-    ratio: np.ndarray,
+    ratio: Union[str, Callable[[np.ndarray], np.ndarray], np.ndarray],
     ratio_uncert: np.ndarray,
     ax: matplotlib.axes.Axes,
     **kwargs: Any,
@@ -456,7 +456,7 @@ def plot_ratio(
 
 def plot_pull(
     _hist: hist.BaseHist,
-    pulls: np.ndarray,
+    pulls: Union[str, Callable[[np.ndarray], np.ndarray], np.ndarray],
     ax: matplotlib.axes.Axes,
     bar_kwargs: Dict[str, Any],
     pp_kwargs: Dict[str, Any],
@@ -505,10 +505,10 @@ def plot_pull(
 
 def _plot_ratiolike(
     self: hist.BaseHist,
-    other: Union[hist.BaseHist, Callable[[np.ndarray], np.ndarray]],
+    other: Union[hist.BaseHist, Callable[[np.ndarray], np.ndarray], np.ndarray, str],
     likelihood: bool = False,
     *,
-    ax_dict: "Optional[Dict[str, matplotlib.axes.Axes]]" = None,
+    ax_dict: Optional[Dict[str, matplotlib.axes.Axes]] = None,
     view: Literal["ratio", "pull"],
     fit_fmt: Optional[str] = None,
     **kwargs: Any,
@@ -676,7 +676,7 @@ def get_center(x: Union[str, int, Tuple[float, float]]) -> Union[str, float]:
 def plot_pie(
     self: hist.BaseHist,
     *,
-    ax: "Optional[matplotlib.axes.Axes]" = None,
+    ax: Optional[matplotlib.axes.Axes] = None,
     **kwargs: Any,
 ) -> Any:
```

src/hist/plot.py
```diff
@@ -214,185 +262,408 @@ def plot2d_full(
...
-        if "alpha" in pp_kwargs:
-            pp_kwargs["alpha"] *= np.power(0.618, i)
+    # Computation and Fit
+    hist_values = self.values()
+
+    main_ax_artists: MainAxisArtists  # Type now due to control flow
+    if callable(other) or isinstance(other, str):
+        if isinstance(other, str):
+            if other in {"gauss", "gaus", "normal"}:
+                other = _construct_gaussian_callable(self)
```
matthewfeickert, line 597: @henryiii I'm trying to change `_plot_ratiolike`'s `other` arg to be type `Union[hist.BaseHist, Callable[[Union[np.ndarray, float]], np.ndarray], str]` but this is now causing `mypy` to error in several places like here with things like

```
error: Incompatible types in assignment (expression has type "Callable[[ndarray, float, float, float], ndarray]", variable has type "Union[BaseHist, Callable[[Union[ndarray, float]], ndarray], str]")
```

Why is `Callable[[ndarray, float, float, float], ndarray]` not compatible with `Callable[[Union[ndarray, float]], ndarray]`? As `Callable` wants a list of types for the first argument it would seem that this should be fine. :?

henryiii: `Callable[[Union[ndarray, float]], ndarray]` is `f(_: ndarray) -> ndarray` or `f(_: float) -> ndarray`.

While `Callable[[ndarray, float, float, float], ndarray]` is f(_: ndarray, _1: float, _2: float, _3: float) -> ndarray`.

I don't see how those are compatible at all.

src/hist/plot.py
matthewfeickert: > Are you sure?

Yes.

henryiii: Ahh, maybe this is never called by `hist.plot_ratio`, so it will truly always get an array. Maybe keeping the names distinct if one does not call the other would make it less confusing? `h.thing()` should call `hist.plot.thing()` or the names should not match.

henryiii: `plot_array_ratio`?

src/hist/plot.py
matthewfeickert: Okay, when presented like this they're obviously not. Previous `mypy` errors for Callable seem to be giving confusing results IMO, but looking again at https://docs.python.org/3/library/typing.html#callable Callable requires you to specify all types of all args if you're not using `...` so I'm guessing that this instead needs to be something ugly looking in terms of Unions.

src/hist/plot.py
matthewfeickert: I can parameterize the tests though to show this for all available `str` options for both `plot_pull` and `plot_ratio`

src/hist/plot.py
henryiii: I'd guess Union of Callables would be needed to cover defaulted arguments? This is a rather weird case since normally you call the callable with `f(x)` (in which case, you just list one) _OR_ `f(a,b,c,d)`.

henryiii: Ahh, it's due to the return type. Maybe it should be the minimum type `Callable[[np.ndarray], np.ndarray]` instead?

src/hist/plot.py
matthewfeickert: Added test.

src/hist/plot.py
matthewfeickert: Thanks &mdash; I didn't know this! Changed this over for all cases where I was using `_hist` to avoid using `hist`.

src/hist/plot.py
matthewfeickert: That will work :+1:  and I guess as that works in `mypy` that my attempts to reflect that the differences were confused.

src/hist/basehist.py
```diff
@@ -401,20 +403,45 @@ def plot2d_full(
 
         return hist.plot.plot2d_full(self, ax_dict=ax_dict, **kwargs)
 
+    def plot_ratio(
+        self,
+        other: "Union[hist.BaseHist, Callable[[np.ndarray], np.ndarray], str]",
```
matthewfeickert, line 408: This is quoted again so that I could add `hist.BaseHist` to the types.

src/hist/plot.py
```diff
@@ -214,185 +262,408 @@ def plot2d_full(
...
+        model_values + model_uncert,
+        **ub_kwargs,
+    )
+
+    return FitResultArtists(line, errorbars, uncertainty_band)
+
+
+def plot_ratio(
+    __hist: hist.BaseHist,
+    ratio: np.ndarray,
```
matthewfeickert, line 367: > FYI, this is the changes I tried; no fixes but added strings / callables and removed quotes in plot.py (since matplotlib is available there) (in .patch form if you want to apply them):

Thanks for the patch as an example. :+1: I had already gotten there when I saw this, but they are very similar with the only exception being that in `plot.py` `plot_ratio` and `plot_pull` should use `np.ndarray` and not `Union[str, Callable[[np.ndarray], np.ndarray], np.ndarray]` as at this point these are both just plotting the already calculated ratios or pulls onto the subplot axis.

src/hist/basehist.py
matthewfeickert: (though maybe I'm being dumb and there is an obvious way to use a class type inside of itself without getting a circular import error)

matthewfeickert: Yes I was being dumb. This was already solved by Henry with

```
T = TypeVar("T", bound="BaseHist")
```

:+1:

src/hist/plot.py
matthewfeickert: > Ahh, maybe this is never called by `hist.plot_ratio` with identical arguments, so it will truly always get an array and it's my fault for adding `str` here. Maybe keeping the names distinct if one does not call the other would make it less confusing? `h.thing()` should call `hist.plot.thing(h)` or the names should not match.

Ah sorry, it seems that I'm missing some of your replies. I've renamed these in `plot.py` to `plot_ratio_array` and `plot_pull_array`.

src/hist/basehist.py
```diff
@@ -401,20 +403,45 @@ def plot2d_full(
 
         return hist.plot.plot2d_full(self, ax_dict=ax_dict, **kwargs)
 
+    def plot_ratio(
+        self,
+        other: Union[T, Callable[[np.ndarray], np.ndarray], str],
```
henryiii, line 408:
```suggestion
        other: Union["hist.BaseHist", Callable[[np.ndarray], np.ndarray], str],
```
You don't need a TypeVar here, you mostly use that to pin something going out to match what's going in. Strings are fine (and this is part of why 3.7 has annotations as strings).

matthewfeickert: SGTM. I have no also realized (doh!) that you can quote single types and not the whole collection of types.

henryiii (approved):
LGTM! Thanks! I'll likely add your example into the docs.

matthewfeickert:
> LGTM! Thanks! I'll likely add your example into the docs.

Awesome. Thanks for being so helpful and patient on this PR @henryiii &mdash; truly appreciate all of your help and guidance here as it ended up being highly educational for me.

eduardo-rodrigues:
> > an argument "gauss" was introduced recently for the most obvious/common use case of the pull plot, so that the user does not have to define that `pdf` function. Could you not do the same fo the ratio plot? That's not the case in the example above.
> 
> @eduardo-rodrigues This is already supported. `hist_1.plot_ratio("normal")` will work (or `"gauss"` or `"gaus"`), but I just didn't show it as such in the example as `plot_ratio` and `plot_pull` now share all the logic internals and only differ in what is plotted in the subplot — the main axis plot logic is the same for both.
> 
> > That's the current design, let's not change that right now. Maybe open a discussion
> 
> @andrzejnovak I would second @henryiii's suggestion here.

Thanks @matthewfeickert. I had not appreciated that. I will comment further on the related thoughts from Henry ...

eduardo-rodrigues:
> I wasn't all that fond of the evaluatable string argument; I would prefer "normal" Python be used, but we should be consistent. I'd be okay to provide a "gauss" function somewhere (hist.gauss or better yet hist.functions.gauss, etc), then remove the strings option entirely, or to only allow "gauss/gaus" and not arbitrary evaulatable strings. But I might be in the minority there too. Let's be consistent, though.
> 
> @andrzejnovak That's the current design, let's not change that _right_ now. Maybe open a discussion? The problem is that these tend to get added directly in the function call, so passing a bunch of dicts is potentially more difficult than passing prefixed keywords (for the record, I think I'm on your side, though). I think you can find examples of both types of API in matplotlib. We could add `something_kw` and combine them, perhaps, etc.

And I tend to agree with you that something more pythonic is preferable, like a callable and a Gaussian function provided by default, for example. Here I'm simply thinking about the user since 99.99% of the time one calls a pull, that's for a Gaussian check (bias and correctness of the error estimation). In that sense, and given that the package is still rather young, I would see no issue in your removing the strings option eventually, TBH. At this point in time, while we don't get there, some consistency is good, as you also say. Happy to continue the discussion in a ... Discussion :-).

PR #168, opened 2021-03-24
feat: fit_fmt and fix some formatting issues

### LINKED ISSUE #159: [FEATURE] Remove fit result summary from pull plot legend as default
**Describe the problem, if any, that your feature request is related to**

PR #149 added placing information from the fit result into the legend for pull plots by default. The example currently in the docs for `plot_pull` now generates this

```python
import hist
import numpy as np
import matplotlib.pyplot as plt
from uncertainties import unumpy as unp


def pdf(x, a=1 / np.sqrt(2 * np.pi), x0=0, sigma=1, offset=0):
    exp = unp.exp if a.dtype == np.dtype("O") else np.exp
    return a * exp(-((x - x0) ** 2) / (2 * sigma ** 2)) + offset


if __name__ == "__main__":
    np.random.seed(0)

    plt.figure(figsize=(10, 8))

    hist_1 = hist.Hist(
        hist.axis.Regular(
            50, -5, 5, name="X", label="x [units]", underflow=False, overflow=False
        )
    ).fill(np.random.normal(size=1000))

    ax, pull_ax = hist_1.plot_pull(
        pdf,
        eb_ecolor="steelblue",
        eb_mfc="steelblue",
        eb_mec="steelblue",
        eb_fmt="o",
        eb_ms=6,
        eb_capsize=1,
        eb_capthick=2,
        eb_alpha=0.8,
        fp_c="hotpink",
        fp_ls="-",
        fp_lw=2,
        fp_alpha=0.8,
        bar_fc="royalblue",
        pp_num=3,
        pp_fc="royalblue",
        pp_alpha=0.618,
        pp_ec=None,
        ub_alpha=0.2,
    )
    ax.get_figure().savefig("pull_plot.png")
```
pull_plot

To my eye, having the fit results on the legend as they are now is not a good idea. It clutters it up and probably shouldn't be on by default. It would be better if it was off by default and required more configuration.

**Describe the feature you'd like**

Have the fit results not be added to the legend with no formatting by default. Have the default be no fit information added to the legend and make it configurable.

### LINKED ISSUE #167: [BUG] Plot artifact in baseline test images
**Describe the bug**

This was found while making test images for PR #161 

The baseline test image for pull plots that was introduced in PR #149 has an axis artifact that is viewable in the pull plot on the sides (it additionally suffers from axis problems from being plotted on subplots).

[](https://github.com/scikit-hep/hist/blob/84a5d80e54ede36b9268e97b0fb545ca68f18957/tests/baseline/test_image_plot_pull.png)

**Steps to reproduce**

```
$ pip list | grep hist
boost-histogram               1.0.0
hist                          2.2.1
histoprint                    2.0.0
```

The following example is trying to replicate most of 

https://github.com/scikit-hep/hist/blob/e12f7485f823931ab1298a0ce2ee55f78b0add4d/tests/test_plot.py#L604-L625

```python
# example.py
import matplotlib.pyplot as plt
import numpy as np

from hist import Hist, axis


def pdf(x, a=1 / np.sqrt(2 * np.pi), x0=0, sigma=1, offset=0):
    return a * np.exp(-((x - x0) ** 2) / (2 * sigma ** 2)) + offset


if __name__ == "__main__":
    np.random.seed(42)

    hist_1 = Hist(
        axis.Regular(
            50, -4, 4, name="S", label="s [units]", underflow=False, overflow=False
        )
    ).fill(np.random.normal(size=1000))
    hist_2 = Hist(
        axis.Regular(
            50, -5, 5, name="S", label="s [units]", underflow=False, overflow=False
        )
    ).fill(np.random.normal(size=10000))

    main_ax, pull_ax = hist_1.plot_pull(pdf)
    fig.savefig("debug_hist_1.png")

    main_ax, pull_ax = hist_2.plot_pull(pdf)
    fig.savefig("debug_hist_2.png")
```

| `debug_hist_1.png` | `debug_hist_2.png` |
| -- | -- |
| debug_hist_1 | debug_hist_2 |

It can be seen that `debug_hist_2.png` (range: `[-5, 5]`) doesn't exhibit the behavior though `debug_hist_1.png` (range: `[-4, 4]`) does, and if you run with a range of `[-6, 6]` it will exhibit the behavior as well.

From the reproduced bug below I think this is due to the axis from the main axis being drawn over but the edges of it sticking out. I'm not sure though, and would defer to someone more expert in `matplotlib`. If you look at the artifact closely it appears to be the right hand side of the range limits (in this case `4` &mdash; it is easier to see with a range of `6`).

henryiii:
I think this fixes #167.

test_image_plot_pull

Also closes #159.


matthewfeickert:
Awesome! Thanks for being super speedy @henryiii. :+1: Also thanks to this I now have learned about [`plt.setp`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.setp.html).

src/hist/plot.py
```diff
@@ -310,12 +314,14 @@ def func(
     # error bar keyword arguments
     eb_kwargs = _filter_dict(kwargs, "eb_")
     eb_kwargs.setdefault("label", "Histogram Data")
+    eb_kwargs.setdefault("fmt", "o")
```
matthewfeickert, line 317: Yeah, I agree that the `"o"` maker looks much better visually then the previous defaults.

tests/test_plot.py
```diff
@@ -620,6 +620,6 @@ def pdf(x, a=1 / np.sqrt(2 * np.pi), x0=0, sigma=1, offset=0):
 
     fig, ax = plt.subplots()
```
matthewfeickert, line 621: As there is still overlap of y-axes on the test plots, can this get changed to 
```suggestion
    https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.setp.html
```
? That then fixes the axis overlap problem
debug_hist_1

matthewfeickert: ...and now with me suggesting actual code and not a URL by accident. sigh.

PR #174, opened 2021-04-05
feat: plo1d for hist(cat,reg)

### LINKED ISSUE #173: [FEATURE]  hist(StrCat, Reg) viz
hist with 1 regular and one StrCat(or even Cat) should visualize as 1d hist.

Currently 
image
and `.plot()` throws an error.

andrzejnovak:
Fixes #173. 

I am kind of expecting to get shouted at for UHI abuse one way or another, but that's what I could piece from the docs on first run. 

image


src/hist/basehist.py
```diff
@@ -301,7 +301,8 @@ def plot(self, *args: Any, **kwargs: Any) -> "Union[Hist1DArtists, Hist2DArtists
         """
         Plot method for BaseHist object.
         """
-        if self.ndim == 1:
+        len_cat = np.sum([ax.traits.discrete for ax in self.axes])
+        if self.ndim == 1 or (self.ndim == 2 and len_cat == 1):
```
henryiii, line 305: I'm not sure I like this. It's also valid to make a 2D plot where one axis is categorical (or two, for that matter). I think it would be more predictable if `.plot()` on a 2D histogram always made a 2D plot. But I could be convinced. What do other people think? @nsmith-, @matthewfeickert, perhaps?

henryiii (commented):
Mostly looks good!

src/hist/basehist.py
andrzejnovak: What @nsmith- 's coffea API does is have overlay keyword to control this for non-discrete axes, but if think the number of 2d histograms of type (sample, quantity) is gonna be higher than (sample, sample) at least if one of the axes is StrCat (or just Cat)

src/hist/basehist.py
```diff
@@ -320,7 +321,16 @@ def plot1d(
 
         import hist.plot
 
-        return hist.plot.histplot(self, ax=ax, **_proc_kw_for_lw(kwargs))
+        if self.ndim == 1:
+            return hist.plot.histplot(self, ax=ax, **_proc_kw_for_lw(kwargs))
+        else:
+            cat_ax = [ax for ax in self.axes if ax.traits.discrete][0]
+            cat_ix = [i for i, ax in enumerate(self.axes) if ax.traits.discrete][0]
```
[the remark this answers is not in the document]
andrzejnovak: 👍 actually shouldn't need cat_ix, about a day later I've found the dict slices can be entered by axis name

src/hist/basehist.py
nsmith-: I suppose in coffea its either `plot1d` or `plot2d`. But indeed if `overlay: Optional[str]` is a parameter, one can take that to decide if a 2D hist is a 2D plot or a set of 1D plots with the overlay axis as specified.

nsmith-: Another option might be to follow pandas style, where the hist would have to be turned into some kind of `HistGroupBy` object before plotting, e.g. `h.groupby("region").plot()`.

andrzejnovak: tbh since `hist` does `plot1d()` and `plot2d()` as well and `plot()` is just a two char shortcut, this is a trifle.

@nsmith- can you think of any other use for some kind of groupby fcn than this? I don't think it's worth fiddling with just for this

nsmith-: Not yet, probably you are right to leave it as-is for now.

andrzejnovak:
@henryiii should I try to make sourcery happy? otherwise this is ready for review I think

src/hist/basehist.py
```diff
@@ -320,7 +325,21 @@ def plot1d(
...
+            else:
+                (cat_ax,) = [ax for ax in self.axes if ax.traits.discrete]
+            if cat_ax.traits.discrete:
+                cats = [cat_ax.value(i) for i in range(len(cat_ax))]
+            else:
+                cats = [bh.loc(cat_ax.value(i)) for i in range(len(cat_ax))]
+            d1hists = [self[{cat_ax.name: cat}] for cat in cats]
+            return hist.plot.histplot(
+                d1hists, ax=ax, label=cats, **_proc_kw_for_lw(kwargs)
+            )
```
henryiii, line 342: Sourcery probably wants this to be "guard" style, where you remove the else and unindent the block. I'm mildly in favor, as less indentation is nicer, and I'm used to guard style from Ruby (where it is quite common and tied into the syntax a bit more).

src/hist/basehist.py
```diff
@@ -297,12 +297,16 @@ def show(self, **kwargs: Any) -> Any:
...
+    ) -> "Union[Hist1DArtists, Hist2DArtists]":
         """
         Plot method for BaseHist object.
         """
-        if self.ndim == 1:
-            return self.plot1d(*args, **kwargs)
+        _has_categorical = np.sum([ax.traits.discrete for ax in self.axes]) == 1
+        _project = _has_categorical or overlay is not None
+        if self.ndim == 1 or (self.ndim == 2 and _project):
+            return self.plot1d(*args, overlay=overlay, **kwargs)
```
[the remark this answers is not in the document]
henryiii: @scikit-hep Does anyone else have an opinion? Should `plot()` be `plot1d` for 1D histograms and `plot2d` for 2D histograms, or should it automatically toggle to `plot1d` for 2D histograms that have a single category axis?

henryiii:
> should I try to make sourcery happy

Only if it wants a sensible change (not always). If it's just the "guard" style if, I mildly like that, so if you are neutral, I'd go with that.

src/hist/basehist.py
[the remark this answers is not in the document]
nsmith-: Since `overlay` is an option here I think I'm in favor of dropping `_has_categorical` condition as well.

src/hist/basehist.py
```diff
@@ -320,7 +325,21 @@ def plot1d(
 
         import hist.plot
 
-        return hist.plot.histplot(self, ax=ax, **_proc_kw_for_lw(kwargs))
+        if self.ndim == 1:
+            return hist.plot.histplot(self, ax=ax, **_proc_kw_for_lw(kwargs))
+        else:
+            if overlay is not None:
+                (cat_ax,) = [ax for ax in self.axes if ax.name == overlay]
```
henryiii, line 332:
```suggestion
            if overlay is not None:
                cat_ax = self.axes[overlay]
```

Will also support integer specification.

src/hist/basehist.py
```diff
@@ -320,7 +325,21 @@ def plot1d(
...
+            else:
+                (cat_ax,) = [ax for ax in self.axes if ax.traits.discrete]
```
henryiii, line 334:
```suggestion
                (cat_ax,) = (ax for ax in self.axes if ax.traits.discrete)
```

No need to iterate twice.

src/hist/basehist.py
```diff
@@ -320,7 +325,21 @@ def plot1d(
...
+            if cat_ax.traits.discrete:
+                cats = [cat_ax.value(i) for i in range(len(cat_ax))]
+            else:
+                cats = [bh.loc(cat_ax.value(i)) for i in range(len(cat_ax))]
```
henryiii, line 338:
```suggestion
            cats = list(cat_ax if cat_ax.traits.discrete else cat_ax.centers())
```

henryiii: FYI, to be PlottableHistogram compliant, you'd want to take the averages of the tuples returned by list(cat_ax) if this is not discrete. But we are hist, we know we have more than that.  Also, please use `hist.loc` instead of `bh.loc` if you do need loc inside hist. :)

src/hist/basehist.py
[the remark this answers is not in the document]
andrzejnovak: It's more typing in total without this assumption. The question is, is it really confusing to assume this? My go-to expectation for just `plot` is 1D.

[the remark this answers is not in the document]
eduardo-rodrigues: In general I do not mind to type 1-2-3 extra characters if that's more explicit and/or clearer. But this is a shortcut, as you say, and one that is likely to be used extensively, hence OK. I tend to agree to expect just `plot` to be 1D.

tests/test_plot.py
```diff
@@ -623,3 +623,31 @@ def pdf(x, a=1 / np.sqrt(2 * np.pi), x0=0, sigma=1, offset=0):
...
+    """
+    Test plot() by comparing against a reference image generated via
+    `pytest --mpl-generate-path=tests/baseline`
+    """
+
+    np.random.seed(42)
+
+    h = Hist(
+        axis.Regular(10, 0, 10, name="variable", label="variable"),
+        axis.StrCategory("", name="dataset", growth=True),
```
henryiii, line 639: Does this support being used with unnamed axes?

src/hist/basehist.py
```diff
@@ -312,6 +316,7 @@ def plot1d(
         self,
         *,
         ax: "Optional[matplotlib.axes.Axes]" = None,
+        overlay: "Optional[str]" = None,
```
henryiii, line 319:
```suggestion
        overlay: "Optional[Union[str, int]]" = None,
```

src/hist/basehist.py
```diff
@@ -320,7 +325,15 @@ def plot1d(
...
 
-        return hist.plot.histplot(self, ax=ax, **_proc_kw_for_lw(kwargs))
+        if self.ndim == 1:
+            return hist.plot.histplot(self, ax=ax, **_proc_kw_for_lw(kwargs))
+        if overlay is not None:
+            cat_ax = self.axes[overlay]
+        else:
+            (cat_ax,) = (ax for ax in self.axes if ax.traits.discrete)
+        cats = cat_ax if cat_ax.traits.discrete else np.arange(len(cat_ax.centers))
+        d1hists = [self[{cat_ax.name: cat}] for cat in cats]
```
henryiii, line 335:
```suggestion
        if overlay is None:
            (overlay,) = (i for i, ax in enumerate(self.axes) if ax.traits.discrete)
        assert overlay is not None
        cat_ax = self.axes[overlay]
        cats = cat_ax if cat_ax.traits.discrete else np.arange(len(cat_ax.centers))
        d1hists = [self[{overlay: cat}] for cat in cats]
```

Maybe something like this would support unnamed axes? The assert is only needed if MyPy needs it (I think it does).

src/hist/basehist.py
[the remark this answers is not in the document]
henryiii: Okay, I'm fine with keeping it since it's a shortcut already, unless anyone is in disagreement.

src/hist/basehist.py
henryiii: I'm not really sure why there are so many quotes on types, but we'll drop Python 3.6 after the next release anyway, so the quotes will go away then.

tests/test_plot.py
andrzejnovak: I'll add it, I think before the changes you proposed it did not

henryiii: Okay, after that, I think this is ready.

henryiii:
@all-contributors please add @andrzejnovak for code

andrzejnovak:
@henryiii looks done then, let me know if you'd like me to squash the commits

henryiii:
Nah, that's fine, I usually squash and merge unless there's a need to leave more than one comment (like a moved and changed file, or a set of sequential steps).

henryiii (approved):
Looks good to me!

henryiii:
I'll fill in the changelog later, to keep the PRs from colliding.

PR #176, opened 2021-04-07
feat: Add frequentist coverage intervals module

matthewfeickert:
Add frequentist coverage intervals as a module (based off those added by @nsmith- to `coffea`) which will be used in PR #161.

Preview of relevant changes to docs: https://hist--176.org.readthedocs.build/en/176/reference/hist.html#module-hist.intervals

## Suggested squash and merge message

```
* Add intervals module to provide frequentist coverage interval support
* Add Literal to hist.typing
```


src/hist/intervals.py
```diff
@@ -0,0 +1,111 @@
...
+        values: Sum of weights.
+        variances: Sum of weights squared.
+        coverage: Central coverage interval.
+          Default is one standard deviation, which is roughly ``0.68``.
+
+    Returns:
+        The Poisson central coverage interval.
+    """
+    # Parts originally contributed to coffea
+    # https://github.com/CoffeaTeam/coffea/blob/8c58807e199a7694bf15e3803dbaf706d34bbfa0/LICENSE
```
matthewfeickert, line 40: @lgray @nsmith- @eduardo-rodrigues moving discussion here now.

henryiii:
I don't think I need to list "add tests for intervals module" to the squashed version, that should be implied by adding the module. :)

src/hist/intervals.py
henryiii: @nsmith-, I'd like at least your sign-off on this, just to be safe. :)

matthewfeickert:
> I don't think I need to list "add tests for intervals module" to the squashed version, that should be implied by adding the module. :)

SGTM. I've revised the PR body. :+1:

src/hist/intervals.py
matthewfeickert: To be clear, I was just adding this as I thought it was required under BSD3, not because I wanted to make things complicated. So as long as @henryiii and the `coffea` team are happy with whatever is here or not here, I'm good.

nsmith-: The attribution looks OK to me

src/hist/intervals.py
```diff
@@ -0,0 +1,129 @@
...
+    if coverage is None:
+        coverage = stats.norm.cdf(1) - stats.norm.cdf(-1)
+    scale = np.empty_like(values)
+    scale[values != 0] = variances[values != 0] / values[values != 0]
+    if np.sum(values == 0) > 0:
+        missing = np.where(values == 0)
+        available = np.nonzero(values)
+        if len(available[0]) == 0:
+            raise RuntimeError(
```
nsmith-, line 49: Inconsistent with docstring. I would suggest `warnings.warn` with a `RuntimeWarning`.

src/hist/intervals.py
```diff
@@ -0,0 +1,129 @@
...
+          Default is ``"poisson"``.
+
+    Returns:
+        The uncertainties for the ratio.
+    """
+    # Note: As return is a numpy ufuncs the type is "Any"
+    with np.errstate(divide="ignore"):
+        ratio = num / denom
+    if uncertainty_type == "poisson":
+        ratio_uncert = np.abs(poisson_interval(ratio, num / np.square(denom)) - ratio)
```
nsmith-, line 121: Maybe some more docs on what this is would be helpful? I am actually not sure what it is, poisson for numerator?

src/hist/intervals.py
```diff
@@ -0,0 +1,129 @@
...
+    elif uncertainty_type == "poisson-ratio":
+        # poisson ratio n/m is equivalent to binomial n/(n+m)
+        ratio_uncert = np.abs(clopper_pearson_interval(num, num + denom) - ratio)
+    else:
```
nsmith-, line 125: Would be nice to add as well the simple propagation of error style uncertainty: https://github.com/CoffeaTeam/coffea/blob/84e805773c1fac32fc79bc9373ec324552371244/coffea/hist/plot.py#L82

nsmith- (changes requested):

src/hist/intervals.py
nsmith-: I looked back at my code to see that indeed it is poisson for numerator.

src/hist/__init__.py
```diff
@@ -5,13 +5,12 @@
 
 from types import ModuleType
 
-from . import accumulators, axis, numpy, storage, tag
+from . import accumulators, axis, intervals, numpy, storage, tag
```
henryiii, line 8: This forces `intervals` to be importable, which in turn forces SciPy to be importable. SciPy (and plotting in general) is not a hard dependency.

henryiii: I'd probably just move the SciPy import into the function(s).

matthewfeickert: Ah yeah, like how the `plot` module does. Will do.

src/hist/intervals.py
matthewfeickert: Sure, but is that needed in this PR to get things up and going for PR #161? That could be added later.

src/hist/intervals.py
matthewfeickert: Henry had suggested the `RuntimeError` and I had forgot to upate the docstring. Why is this a warning though vs. an error? If everything is zero then something is wrong.

src/hist/intervals.py
```diff
@@ -0,0 +1,139 @@
+from typing import Any, Optional
+
+import numpy as np
+
+from .typing import Literal
+
+try:
+    from scipy import stats
+except ImportError:
```
henryiii, line 9:
```suggestion
except ModuleNotFoundError:
```

src/hist/intervals.py
henryiii: Feel free to open as an issue and move forward with this for now. :)

src/hist/intervals.py
nsmith-: It's valid to plot an empty histogram with error bars, certainly if the data is not scaled then you even have a well-defined error bar.  The interpretation of them can be challenging if the data is scaled, however, and there's no way to tell other than the user. Either way it seems we should not throw an exception.

src/hist/intervals.py
matthewfeickert: Do you want me to grab that in `hist.plot` as well? 

https://github.com/scikit-hep/hist/blob/929704218bc9fb24875eafaa1296ab0763e33d72/src/hist/plot.py#L235

I originally had it as `ModuleNotFoundError` but then changed to match.

src/hist/intervals.py
nsmith-: Considering its the ROOT default ratio plot error (`TH1::Divide`) seems we would want that no?

src/hist/intervals.py
henryiii: Sure. I thought you had, was surprised to see it revert. `ModuleNotFoundError` is the more precise (Python 3 only) error to catch.

src/hist/intervals.py
matthewfeickert: > Considering its the ROOT default ratio plot error (`TH1::Divide`) seems we would want that no?

Let me go look, but I'm not sold on this argument. ROOT does plenty of things that are not a good idea.

henryiii: I also do not consider ROOT's warning vs. error behavior to be a design standard for us. :)

Now if there's a valid reason to do this, then why is there a warning? And if not, why not just error now?

henryiii: (If it's valid but rare, warnings can be squelched, while error's can't be)

nsmith-: I didn't add it originally either, but it is valid in some cases. See the PR that added it to coffea: https://github.com/CoffeaTeam/coffea/pull/182

src/hist/intervals.py
matthewfeickert: > It's valid to plot an empty histogram with error bars

Why are you plotting non-existent data? If the histogram is actually empty this doesn't make sense. Link to an example?

matthewfeickert: > But imagine you are dumping 100 region plots to a file

IMO it should crash. It is your responsibility to clean your data.

nsmith-: Ok, its a valid opinion. Just seems at odds with the amount of work this routine does to make up vaguely reasonable error bars in the case where even all but one bin is zero, that it would give up if all are zero. I won't press the issue further then (except to update the docstring)

henryiii: If you expect it to crash, that's what try/except (or using "if" beforehand) is for. This could mask real problems, like all plots being empty because something is misconfigured? Nobody reads logs; warnings are next to useless.

src/hist/intervals.py
matthewfeickert: I'm switching work at the moment, but I'll come back tonight and clean this up and make it more clear.

src/hist/intervals.py
matthewfeickert: > Isn't this poisson interval method used for plain errorbar histos (as well as ratios)?

In practice it could be, but at the moment it is only being used in ratio_uncertainty.

> Just seems at odds with the amount of work this routine does to make up vaguely reasonable error bars in the case where even all but one bin is zero, that it would give up if all are zero

I see what you're saying, but then would you also suggest not allowing any empty values? Or how would you define the cutoff? I don't think that you're pressing anything, I just want to make sure that I understand what your thoughts are here as it is clear that you've thought about this far more than I have.

(Henry I think we're in agreement but if I'm misunderstanding (sorry) please let me know).

src/hist/intervals.py
matthewfeickert: > I didn't add it originally either, but it is valid in some cases.

Looking back at that Issue it is brought up that

> > Ah yeah, fair enough, this would be needed for efficiencies derived from weighted data (which is kind of rare, but definitely happens).

But the specific example isn't really elaborated on. Can you either ELI5 why this is worth doing now, or make a new Issue from this discussion and it can get labeled as a "good first issue"? Again, you've thought about this all more than I have, so it is very possible I'm just not seeing something incredibly obvious to you.

nsmith-: I thought it would be convenient to copy it as the others, since it isn't much work. But also fine to put it off to later if you prefer. I can make the issue

src/hist/intervals.py
nsmith-: The all zero case is the only situation when there really isn't enough info to do something reasonable so its fine. LGTM

src/hist/intervals.py
matthewfeickert: > Poke me when ready

sorry I missed this comment. poke.

nsmith- (approved):

henryiii (approved):
LGTM, thanks!

PR #190, opened 2021-04-12
docs: Revise plot_ratio and plot_pull API examples in user guide

### LINKED ISSUE #188: [DOCS] Revise plot_ratio and plot_pulll docs examples in Plots notebook
## How do you want to improve the docs

There's a few things in the existing docs example notebooks that aren't correct. For example, the original `plot_pull` example has the following line

```python
ax, pull_ax = hh.plot_pull(
    pdf,
    eb_ecolor="steelblue",
    eb_mfc="steelblue",
    eb_mec="steelblue",
    eb_fmt="o",
    eb_ms=6,
    eb_capsize=1,
    eb_capthick=2,
    eb_alpha=0.8,
    fp_c="hotpink",
    fp_ls="-",
    fp_lw=2,
    fp_alpha=0.8,
    bar_fc="royalblue",
    pp_num=3,
    pp_fc="royalblue",
    pp_alpha=0.618,
    pp_ec=None,
    ub_alpha=0.2,
)
```

which is no longer correct, as `plot_pull` and `plot_ratio` return artist tuples now and not axes. The `plot_pull` example in the Plot Ratio examples could probably just be moved up and then the `plot_ratio` examples could have objects assigned to the returns so the artist tuples aren't printed out in the notebook. So something like

```python
artists = hist_1.plot_ratio(
    hist_2,
    rp_ylabel=r"Ratio",
    rp_num_label="hist1",
    rp_denom_label="hist2",
    rp_uncert_draw_type="bar",
);
```

## Where is the typos and defects

https://github.com/scikit-hep/hist/blob/7e8c76187899dff659d97e1eef3a9469b4966a8b/docs/user-guide/notebooks/Plots.ipynb

## What contents do you expect

Revisions as oultined above.

matthewfeickert:
Resolves #188

This PR revised the Plot notebook in the user's guide to note the return structure of both `plot_ratio` and `plot_pull` is a tuple of matplotlib artists and to expand on the use of both callables and `str` aliases in both APIs.

Relevant section of RTD preview: https://hist--190.org.readthedocs.build/en/190/user-guide/notebooks/Plots.html#Via-Plot-Pull

### DIFF
--- a/docs/user-guide/notebooks/Plots.ipynb
+++ b/docs/user-guide/notebooks/Plots.ipynb
@@ -200,19 +200,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from uncertainties import unumpy as unp\n",
-    "\n",
-    "\n",
     "def pdf(x, a=1 / np.sqrt(2 * np.pi), x0=0, sigma=1, offset=0):\n",
-    "    exp = unp.exp if a.dtype == np.dtype(\"O\") else np.exp\n",
-    "    return a * exp(-((x - x0) ** 2) / (2 * sigma ** 2)) + offset"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "*(The uncertainty is non-significant as we filled a great quantities of observation points above.)*"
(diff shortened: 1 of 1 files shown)


henryiii (approved):
LGTM, thanks!

PR #194, opened 2021-04-13
chore: drop Python 3.6

henryiii:


src/hist/basehist.py
```diff
@@ -352,10 +348,10 @@ def plot(
     def plot1d(
         self,
         *,
-        ax: "Optional[matplotlib.axes.Axes]" = None,
-        overlay: "Optional[Union[str, int]]" = None,
+        ax: matplotlib.axes.Axes | None = None,
+        overlay: str | int | None = None,
         **kwargs: Any,
-    ) -> "Hist1DArtists":
+    ) -> Hist1DArtists:
```
henryiii, line 354: @matthewfeickert This is an example of Python 3.7+ syntax for type annotations (I noticed Python 3.6 syntax in a recent PR of yours). You can use isort to inject `from __future__ import annotations` to all your source files.

matthewfeickert: Ah thanks! That notation is really slick. :)

PR #235, opened 2021-06-11
docs: link CONTRIBUTING.md to documentation

amangoel185:
Links CONTRIBUTING.md to the [official documentation](https://hist.readthedocs.io/).

- eliminates the need for having a separate [`development.rst`](https://github.com/scikit-hep/hist/blob/main/docs/development.rst)
- linking done by referring to similar functions implemented in [boost-histogram](https://github.com/scikit-hep/boost-histogram/blob/54beb7f19a288b07be8a4ec48819bacd1759bbef/docs/conf.py)


henryiii:
Feel free to make this non-draft when you are ready for review and merge (looks good AFAICT?)

amangoel185:
Sure, looks good to me as well! :)

I was just wondering whether we want to keep the contribution page title as **Setting up a development environment** or something else like **Development** or **Contributing**?

henryiii:
Sure, you can change it - the "title" was implicitly Contributing before due to the file name. Mark as ready for review when ready (currently draft).

henryiii:
Let's call it Contributing, since that's the name of the file. And it's really only needed for contributing to hist; you might be "developing" a different package, then you don't need this page, but just a normal pip install.

amangoel185:
Alright, that makes sense! :)

PR #241, opened 2021-06-19
fix: reprs had missing information

### LINKED ISSUE #200: [BUG] Repr issues
The integer repr is missing name/label, and label gets repeated even if it's just picking it up from name for the other axes, like Regular.

amangoel185:
Fixes #200

### DIFF
--- a/src/hist/axis/__init__.py
+++ b/src/hist/axis/__init__.py
@@ -141,7 +141,7 @@ class Boolean(AxesMixin, bha.Boolean, family=hist):
         self.label: str = label
 
 
-class Variable(bha.Variable, AxesMixin, family=hist):
+class Variable(AxesMixin, bha.Variable, family=hist):
     __slots__ = ()
 
     def __init__(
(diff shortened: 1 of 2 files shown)


henryiii:
Could you add a test for the repr to make this doesn't happen again?

amangoel185:
Sure

amangoel185:
Sure, I'll do that.

amangoel185:
I've added the tests, @henryiii. You can merge it if it looks good to you.

src/hist/axis/__init__.py
```diff
@@ -212,8 +212,8 @@ def __init__(
         *,
         name: str = "",
         label: str = "",
-        growth: bool = False,
         metadata: Any = None,
+        growth: bool = False,
```
henryiii, line 216: Not sure why these got reordered, but due to the "*" above, order does not affect users, so it's safe. :)

henryiii (approved):

src/hist/axis/__init__.py
amangoel185: Oh, I reordered them since I saw that order in `Regular`.

PR #243, opened 2021-06-22
feat: axis sort

### LINKED ISSUE #222: [FEATURE] Reorder categorical axes
I've hit this while implementing https://github.com/scikit-hep/mplhep/issues/303 and realizing the results can be funny if the categorical axis is set with `growth=True`. I think it makes more sense to order on the object, than at plotting time.  API could look like 
```
h.sort('axis_name', key=list | fcn as passed to sorted())
```

andrzejnovak:
When completed, closes #222 

To implement:
- sort by passing index
- sort by passing lambda to sorted
- add tests

@henryiii could you take a look? I had to abuse some design decisions of hist to get this work, so maybe you can recommend better workarounds. 
- AxisNamedTuple (and I assume axes) as well cannot be assigned to, so I had to recreate the sorted axis
- Couldn't easily copy all meta/traits from old axis to new - hence the helper with inspect (name/label don't seem to be part of traits/metadata) - also cannot create axis and edit metadata

Here is a couple of examples 
2D - yax
image
2D - xax
image
1D
image
3D - sort on not-projected axis
image

### DIFF
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -1,5 +1,5 @@
 import functools
-import inspect 
+import inspect
 import operator
 import typing
 import warnings
@@ -229,7 +229,6 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
         total_data = tuple(args) + tuple(data)  # Python 2 can't unpack twice
         return super().fill(*total_data, weight=weight, sample=sample, threads=threads)
 
-    
     def sort(
         self: T,
         axis: Union[int, str],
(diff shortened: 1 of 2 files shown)


src/hist/basehist.py
```diff
@@ -210,6 +211,45 @@ def fill(
         total_data = tuple(args) + tuple(data)  # Python 2 can't unpack twice
```
henryiii, line 211: Totally unrelated, but we should unpack twice, hist has never supported Python 2... :)

henryiii:
Would https://github.com/scikit-hep/boost-histogram/pull/577 make this easier? `h[{i: sorted(h.axis[i]})]` would do it, I think? (I'm applying hist's UHI+ on top of boost-histogram's PR in this example)

henryiii:
> Couldn't easily copy all meta/traits from old axis to new

I think, post refactor, this is now just copying the `__dict__` from the old to the new.

```pycon
>>> h = Hist.new.Reg(10, 0, 1, name="hi", label="ho", __metadata__ = {"this": "that"} ).Double()
>>> h.axes[0].this = "that"
>>> h.axes[0].__dict__
{'name': 'hi', 'label': 'ho', 'this': 'that'}
```

andrzejnovak:
Ok yeah, arguably scikit-hep/boost-histogram#577 could make this entire API redundant, though it might still be worth having as a shorthand for the 2 most obvious use cases I see - sort by label, sort by integral.

Btw the bh draft doesn't work for StrCat, I guess you are aware with the `bh.loc` comment. 

Any ETA on that draft? This sorting histograms question comes up quite often for me and it's fairly experience breaking to reply with "export to 1Ds and pass as a list to mplhep".

henryiii:
Before PyHEP; but hopefully this week. Probably will be right after the PR above it (scaling).

henryiii:
This is going in as an "experimental" feature in boost-histogram 1.1 - scikit-hep/boost-histogram#577 . It does not add removed bins to the flow bin when a selection is not complete, but that should not matter for this PR, so the warning could be silenced.

henryiii:
Only have a sort by label currently, but I think this is good enough for a first pass?

andrzejnovak:
Great, thanks a lot! It's amazing how much easier this is with the axis picking implemented.

LovelyBuggies:
> Only have a sort by label currently, but I think this is good enough for a first pass?

FYI, do you mean `h.project(*[i[0] for i in sorted(enumerate(h.axes.label), key=lambda x:x[1], reverse=False)])`? Or do we have another shortcut? 

```python
from hist import Hist

h = Hist.new.StrCat(["AB", "BCC", "BC"], name="Y", label="Y-axis").IntCat([1, 4, 2], name="X", label="X-axis").Double()
# sort StrCat
h.sort(0, reverse=True).axes[0]
# sort axes by name
h.project(*[i[0] for i in sorted(enumerate(h.axes.name), key=lambda x:x[1])]) # we may want h.sort_by_names
# sort axes by label
h.project(*[i[0] for i in sorted(enumerate(h.axes.label), key=lambda x:x[1], reverse=False)]) # we may want h.sort_by_labels
```

henryiii:
No, I mean you only have access to the StrCategory label (like `AB`), and not the position (but you could get that by capturing it, so I don't think there's any loss). We don't have a way to sort the axes, but that's likely not needed, either just create the histogram with the axes in a different order, don't rely on the axis order at all, or use project.

Issue #169, opened 2021-03-24
[FEATURE] HistogramStack


henryiii:
A possible design for stacked histograms is the following:

A HistogramStack holds multiple histograms; axes are required to match. Calling `.plot(...)` forwards to mplhep, just like Hist.plot, but with the stack of histograms as a list in the first argument.

To make a stack, you could use `|`, so that:

```python
(h1 | h2 | h3).plot()
```

Would make a stack of histograms and plot it.

eduardo-rodrigues:
@henryiii, just thinking out loud, why a "|" rather than a "&"?

BTW, did you look at some of the functionality that was put into histbook in the past, before it got archived? Stacked histograms were also covered there and one may "steal" ideas from it.

eduardo-rodrigues:
I see, fair enough for the "|", then.

henryiii:
FYI: In Python 3.10, `int | str` is the same as `from typing import Union; Union[int, str]`, and it will work inside isinstance, so you can write `isinstance(x, int | str)` instead of `isinstance(x, (int, str))`.

henryiii:
`hist.stacks.Stack(h1, h2)` would be the explicit constructor. There also should be some way to convert a category axis to a stack. `h.stack("axis1") -> stack`.

LovelyBuggies:
For implicit constructor, aka `s = h1 | h2 | h3`, we probably need to overload `__or__` for both Hist and Stack. Hence, IMO, `s.push(h4)` could be helpful (and `s.pop()` correspondingly).

henryiii:
I'm thinking 95% of users will likely do `h.stack(ax)` and not `Stack(hist1, hist2, ...)`, so I'm not that worried about adding a shortcut for it now. If we do, yes, but let's not add it unless we find a strong need for it. Things like #274 are more pressing.

PR #244, opened 2021-06-28
feat: HistStack for #169

LovelyBuggies:
To solve #169.


.pre-commit-config.yaml
```diff
@@ -57,7 +57,7 @@ repos:
   - id: mypy
     files: ^src
     args: []
-    additional_dependencies: ["numpy>=1.20", "matplotlib>=3.3", "boost-histogram~=1.0.1", "uhi>=0.2.1"]
+    additional_dependencies: ["numpy==1.20", "matplotlib>=3.3", "boost-histogram~=1.0.1", "uhi>=0.2.1"]
```
amangoel185, line 60: is this an intentional change?

henryiii: See #245. Though I think it should be 
```suggestion
    additional_dependencies: ["numpy==1.20.*", "matplotlib>=3.3", "boost-histogram~=1.0.1", "uhi>=0.2.1"]
```

LovelyBuggies: Suggestion commited.

LovelyBuggies:
Maybe with more plotting options here. Any suggestions?

henryiii:
```python
h =  Hist.new.Reg(..., name="x").StrCat(["one", "two"] name="y").Double()

s = h.stack("y")
s = Stack({"one": h[:, "one"], "two":h[:, "two"]})

# Another way?
h1 = h[:, "one"]
h1.name = "one"
h2 = h[:, "two"]
h2.name = "two"
s = Stack(h1, h2)

h.plot() # -> shortcut for h.stack("y").plot()
```

Thought: Could we above a initial memory copy for stacking an existing histogram? May not be better, due to making copy every access?
```
self._data = h
self[x] -> pull x from histogram _OR_ list/dict?
```

LovelyBuggies (commented):
~~We also need some tests for Stack(axes) and Hist.stack(name/idx)~~

Done.

src/hist/basehist.py
```diff
@@ -461,3 +461,26 @@ def plot_pie(
         import hist.plot
 
         return hist.plot.plot_pie(self, ax=ax, **kwargs)
+
+    def stack(self: T, *args: Union[int, str]) -> Any:  # todo: should return Stack here
```
henryiii, line 465:
```suggestion
    def stack(self: T, *args: Union[int, str]) -> "hist.stacks.Stack":
```

And add an `import hist.stacks` at the top. You can avoid a circular import error by not referring to "Stack" directly outside of a function body.

src/hist/basehist.py
```diff
@@ -461,3 +461,26 @@ def plot_pie(
...
+        Returns a stack from a normal histogram axes.
+        """
+        from .stacks import Stack
+
+        if len(args) == 0:
+            raise ValueError("There should be histograms or axes in stack")
+
+        if isinstance(args[0], int):
+            stack_args = [self.axes[i] for i in args]
+            return Stack(*stack_args)
```
henryiii, line 476: I think we should only support one arg, which is the axis to stack _over_. Multiple args would make a ND stack, which we don't want to support, at least for now.

The argument should be the axis to stack over, which should be expanded. For example:

```python
ax = 0
stack_histograms = {i:self[i, ...] for i in range(len(ax[0]))])
return Stack(*stack_histograms)
```

Though I think we also need to keep track of the bin this came from too.

src/hist/basehist.py
LovelyBuggies: Fxied.

src/hist/basehist.py
LovelyBuggies: What's the `ax` in this example?

henryiii: In my case, it was the integer axis number to expand. Though if it was a string axis name, since that is supported natively in both `h[{...}]` and `h.axis[...]`, that should work too. I didn't expand the dict correctly above, though.

henryiii: Updated the example above - it now no longer "remembers" the cell name, though. Ideally, each histogram would also have at least "h.name" set to the original cell "name". Maybe:
```python
for name, h in zip(self.axis[ax], stack_histograms)
    h.name = name
```

henryiii: If `self.axes[i]` is a `StrCategory`, which is most common, then `h.name` will be the category that this histogram came from. If it's an integer, then name is an integer. If it's a Regular/Variable, then "name" is a tuple with two floats, which is weird for something called "name". But I think it's a reasonable tradeoff - having the value is still useful.

LovelyBuggies: Is there any difference between self.axis and self.axes?

henryiii:
Quick summary of changes: axes are not stackable, only histograms. Added a little bit of container-like support so `stack[0]` gets the 0'th histogram in the stack. Added .name which gets set from the category. Did a little parametrization of the tests (more could be done eventually). I'm reserving `overlay=` because I forgot what it was supposed to do. I also used `stack.py` to match `Stack`, and used `self.__class__.__name__` in the repr.

henryiii:
Thanks everyone!

LovelyBuggies:
> Added `.name` which gets set from the category (but currently unused by mplhep or us). 

How to use `.name` and `h.stack()`?

henryiii:
I thought there were some. This is what it looks like:

```python
h = hist.Hist.new.Reg(10,0,1, name="x").StrCategory(["good", "bad"], name="quality").Double()
# Fill with data tagged with quality="good" or "bad"
stack = h.stack("quality")
assert stack[0].name == "good"
assert stack[1].name == "bad"
```

Ideally, calling `stack.plot(); plt.legend()` would produce a legend that labels these with "good" and "bad", but that doesn't happen yet, might need to be added to mplhep. Unless there's a labels keyword or something, which I didn't check for.

Could you add a test like the one above, actually? I'm working on three other features.

henryiii:
Oops, forgot, `StrCat`. Allowing both `StrCat` and `StrCategory` is allowed on the branch I'm working on - and accidentally passing the quick constructor to the Hist as an axis is allowed, too!

LovelyBuggies:
FYI,

image

henryiii:
This is why we write tests! 😳

PR #250, opened 2021-07-06
fix: modify error message and add warning

### LINKED ISSUE #233: [BUG] Hist does not forbid axes names that are reserved keywords for fill
None of these names are legal because they are reserved in the `fill` signature
```python
import hist

h = hist.Hist(
    hist.axis.Regular(10, 0, 10, name="weight"),
    hist.axis.Regular(10, 0, 10, name="sample"),
    hist.axis.Regular(10, 0, 10, name="threads"),
)
h.fill(weight=1, sample=2, threads=3)
```

amangoel185:
Fixes #233

### DIFF
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -94,6 +94,12 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
                 warnings.warn(msg)
                 storage = storage()
             super().__init__(*args, storage=storage, metadata=metadata)  # type: ignore
+
+            disallowed_names = ["weight", "sample", "threads"]
+            if any(item in disallowed_names for item in [ax.name for ax in self.axes]):
+                disallowed_warning = "You have used a disallowed name for axes"
+                warnings.warn(disallowed_warning)
+
             valid_names = [ax.name for ax in self.axes if ax.name]
             if len(valid_names) != len(set(valid_names)):
                 raise KeyError(
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -97,7 +97,9 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
 
             disallowed_names = ["weight", "sample", "threads"]
             if any(item in disallowed_names for item in [ax.name for ax in self.axes]):
-                disallowed_warning = f"{item} is a protected keyword and cannot be used as axis name"
+                disallowed_warning = (
+                    f"{item} is a protected keyword and cannot be used as axis name"
+                )
                 warnings.warn(disallowed_warning)
 
             valid_names = [ax.name for ax in self.axes if ax.name]
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -209,7 +209,9 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
         }
 
         if set(data_dict) != set(range(len(args), self.ndim)):
-            raise TypeError("All axes must be accounted for in fill")
+            raise TypeError(
+                "All axes must be accounted for in fill, you may have used a disallowed name in the axes"
+            )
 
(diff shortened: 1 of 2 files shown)


src/hist/basehist.py
```diff
@@ -94,6 +94,12 @@ def __init__(
                 warnings.warn(msg)
                 storage = storage()
             super().__init__(*args, storage=storage, metadata=metadata)  # type: ignore
+
+            disallowed_names = ["weight", "sample", "threads"]
+            if any(item in disallowed_names for item in [ax.name for ax in self.axes]):
+                disallowed_warning = "You have used a disallowed name for axes"
```
andrzejnovak, line 100:
```suggestion
                disallowed_warning = f"{item} is a protected keyword and cannot be used as axis name"
```

Are we past 3.6 yet?

amangoel185: We'll drop 3.6 after this release.

LovelyBuggies (commented):
Could you add tests (three expection tests)

amangoel185:
Have added the tests, @LovelyBuggies.

tests/test_axis.py
```diff
@@ -51,3 +53,14 @@ def test_axis_flow():
...
+
+def test_axis_disallowed_names():
+
+    with pytest.warns(UserWarning):
+        h = hist.Hist(
+            axis.Regular(10, 0, 10, name="weight"),
+            axis.Regular(10, 0, 10, name="test"),
+            axis.Regular(10, 0, 10, name="threads"),
+        )
+        h
```
LovelyBuggies, line 66: we don't actually need the `h` in the last line.

LovelyBuggies: Change to this:

```python
warns ...:
    h = hist.Hist(axis.Regular(10, 0, 10, name="weight"))
warns ...:
    h = hist.Hist(axis.Regular(10, 0, 10, name="test"))
warns ...:
    h = hist.Hist(axis.Regular(10, 0, 10, name="threads"))
```

LovelyBuggies:
LGTM now.

src/hist/basehist.py
```diff
@@ -94,6 +94,13 @@ def __init__(
                 warnings.warn(msg)
                 storage = storage()
             super().__init__(*args, storage=storage, metadata=metadata)  # type: ignore
+
+            disallowed_names = ["weight", "sample", "threads"]
```
henryiii, line 98:
```suggestion
            disallowed_names = {"weight", "sample", "threads"}
```

Ultra minor, but this is slightly more logically correct. It's technically probably a bit slower, just because sets take longer to construct than lists/tuples unless you are constructing them and testing containership in the same line. Actually checking it is faster, so if you did it multiple times, sets are a clear win. But we aren't talking about a significant time in any case.

src/hist/basehist.py
```diff
@@ -203,7 +210,9 @@ def fill(
         }
 
         if set(data_dict) != set(range(len(args), self.ndim)):
-            raise TypeError("All axes must be accounted for in fill")
+            raise TypeError(
+                "All axes must be accounted for in fill, you may have used a disallowed name in the axes"
```
henryiii, line 214: Eventually I'd only do this if you have an axis with a disallowed name, but this is fine for now.

I'd eventually make the `disallowed_names` a global `DISALLOWED_NAMES: Final[set[str]] = {...}`, then we could see if there is any overlap between these, and only suggest the "you may have used a disallowed name in the axes" part if there is a chance of this - most users seeing this error probably are making normal mistakes, not using one of these special names

src/hist/basehist.py
amangoel185: Alright! That makes sense. :)

src/hist/basehist.py
amangoel185: Oh okay, we can make it global after this.

PR #258, opened 2021-07-07
fix: more testing and label support

henryiii:
- tests: nicer config
- fix: support label and better testing

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -20,7 +20,7 @@ def lint(session):
     session.run("pre-commit", "run", "--all-files", *session.posargs)
 
 
-@nox.session(python=ALL_PYTHONS)
+@nox.session(python=ALL_PYTHONS, reuse_venv=True)
 def tests(session):
     """
     Run the unit and regular tests.
--- a/src/hist/stack.py
+++ b/src/hist/stack.py
@@ -1,11 +1,14 @@
-import sys
 import typing
-from typing import Any, Iterator, List, Tuple, Union
+from typing import Any, Iterator, List, Tuple, TypeVar, Union
 
 from .basehist import BaseHist
 
 if typing.TYPE_CHECKING:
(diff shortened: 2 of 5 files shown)


src/hist/stack.py
```diff
@@ -30,35 +33,48 @@ def __init__(
...
+
+    def plot(self, **kwargs: Any) -> "List[Hist1DArtists]":
         """
         Plot method for Stack object.
         """
-        if overlay is not None:
-            raise NotImplementedError("Currently overlay is not supported")
 
-        if self._stack[0].ndim != 1:
+        import hist.plot
```
henryiii, line 65: Besides piggy-backing on our existing mechanism for replacing this, it also makes it easy to monkeypatch in tests, too!

src/hist/stack.py
```diff
@@ -30,35 +33,48 @@ def __init__(
...
-    def __getitem__(
-        self, val: Union[int, slice]
-    ) -> Union[BaseHist, Tuple[BaseHist, ...]]:
         return self._stack.__getitem__(val)
 
     def __iter__(self) -> Iterator[BaseHist]:
         return iter(self._stack)
 
-    def plot(self, *, overlay: None = None, **kwargs: Any) -> "List[Hist1DArtists]":
+    def __len__(self) -> int:
```
henryiii, line 53: Forgot this earlier.

noxfile.py
```diff
@@ -20,7 +20,7 @@ def lint(session):
     session.run("pre-commit", "run", "--all-files", *session.posargs)
 
 
-@nox.session(python=ALL_PYTHONS)
+@nox.session(python=ALL_PYTHONS, reuse_venv=True)
```
henryiii, line 23: I think this is safe. Let's try it, anyway; it saves quite a few (15-20) seconds for me.

PR #264, opened 2021-07-17
docs: integrate boost-histogram's docs

henryiii:
We should use boost-histograms's docs as a base, and link back to them as well. Just getting started here on QuickStart.


amangoel185:
Establishes base for #265.

Issue #265, opened 2021-07-17
[DOCS] Copy in structure and content from boost-histogram


henryiii:
We should include the contents and use the structure of the boost-histogram docs, just slightly modified to suit Hist usage. Most user's won't want to read both sets of documentation. Started in #155.

PR #270, opened 2021-07-20
docs: add analyses examples to documentation

amangoel185:
Adds on to #265.

### DIFF
--- a/docs/user-guide/analyses.rst
+++ b/docs/user-guide/analyses.rst
@@ -36,7 +36,7 @@ True ``is_valid`` selection, you can use a ``sum``:
 
 .. code:: python3
 
-   h1 = hist[:, True, :: sum]
+   h1 = hist[:, True, sum]
 
 You can expand this example to any number of dimensions, boolean flags,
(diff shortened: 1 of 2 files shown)


docs/user-guide/analyses.rst
```diff
@@ -0,0 +1,42 @@
...
+   bool_ax = hist.axis.Integer(0, 2, underflow=False, overflow=False)
+   run_number_ax = hist.axis.IntCategory([], growth=True)
+
+Now, you can use these axes to create a single histogram that you can
+fill. If you want to get a histogram of all run numbers and just the
+True ``is_valid`` selection, you can use a ``sum``:
+
+.. code:: python3
+
+   h1 = hist[:, True, :: sum]
```
henryiii, line 39:
```suggestion
   h1 = hist[:, True, sum]
```

Technically this is the third item (the action), but you don't have to be explicit, if it's the only think it's still handled correctly as a shortcut.

henryiii (approved):

PR #271, opened 2021-07-20
docs: update the Axes documentation

amangoel185:
Adds on to #265.

- [x] TODO: add images for axes


amangoel185:
@henryiii, could you verify if the updated heading levels seem fine?

henryiii (approved):
I think so - let's go with it and we can tweak later if we need to. :)

PR #266, opened 2021-07-17
feat: Add "efficiency" option to rp_uncertainty_type kwarg

matthewfeickert:
Resolves #195

Add an `"efficiency"` option to `ratio_uncertainty`'s `uncertainty_type` kwarg that acts as an alias for the `poisson-ratio` option. Additionally add an example of making an efficiency plot to the user guide.

PR #278, opened 2021-08-03
fix: Use proper Clopper-Pearson interval for efficiency ratio

### LINKED ISSUE #260: [BUG] Error message when category slicing info isn't present in the axies
**Describe the bug**

If you create a category axis, fill, and then slice it using the list operator when you ask for a category that doesn't exist the error message is rather hard to interpret. Should be improved.

**Steps to reproduce**

This [notebook](https://github.com/iris-hep/opendata-higgs-discovery/blob/main/notebooks/hist-demo-cat-clice-errors.ipynb) shows how this works with `hist` version 2.4.0. A error message might be more like "x, y, and z, are valid in this category axis" or similar.

### LINKED ISSUE #277: [BUG] Uncertainty bands in efficiency ratios go above unity
**Describe the bug**

The error bars produced in the ratio of an efficiency style plot (`hist.plot_ratio(..., rp_uncertainty_type = "poisson-ratio", ...)`)
can extend above unity. This is unexpected and not meaningful if the numerator is a true subset of the denominator.

**Steps to reproduce**

This is observed using `hist==2.4.0`.

Following the example from the [docs](https://hist.readthedocs.io/en/latest/user-guide/notebooks/Plots.html),

```python
hist0 = hist.Hist(hist.axis.Regular(50, -5, 5), underflow = False, overflow = False).fill(np.random.normal(size=1700))
hist1 = hist0.copy() * 0.98

fig = plt.figure(figsize = (10,8))
_ = hist1.plot_ratio(
    hist0,
    rp_ylabel="Efficiency",
    rp_num_label="hist1",
    rp_denom_label="hist0",
    rp_uncert_draw_type="line",
    rp_uncertainty_type="poisson-ratio",
    rp_ylim=[0.9,1.1],
)
```
This produces something like the following,
image

where we see the error bars going above unity.

If I take the same data used in `hist0` and `hist1` above, and produce a `TEfficiency` object in `ROOT`, I get something like the following,
image
which is something more like what I would expect since I believe `hist` is using a Clopper-Pearson interval (default in ROOT).

dantrim:
Fixes #277.

```
* Use the correct Clopper-Pearson interval for an efficiency ratio as for efficiency numerator is subset of denominator
   - Update docstring to reflect this
* Add default values for ratio plot ylabel and ylim if `efficiency` option used
* Correct efficiency plot example in user guide
* Add test to ensure no uncertainties are above 1
```

### DIFF
--- a/src/hist/intervals.py
+++ b/src/hist/intervals.py
@@ -128,7 +128,9 @@ def ratio_uncertainty(
          numerator scaled by the denominator.
          ``"poisson-ratio"`` implements the Clopper-Pearson interval for Poisson
          distributed ``num`` and ``denom``.
-         ``"efficiency"`` is an alias for ``"poisson-ratio"``.
+         ``"efficiency"`` implements the Clopper-Pearson interval for Poisson
+         distributed ``num`` and ``denom``, with ``denom`` assumed to be a
+         strict subset of ``num``.
 
     Returns:
         The uncertainties for the ratio.
@@ -138,9 +140,11 @@ def ratio_uncertainty(
         ratio = num / denom
     if uncertainty_type == "poisson":
         ratio_uncert = np.abs(poisson_interval(ratio, num / np.square(denom)) - ratio)
-    elif uncertainty_type in {"poisson-ratio", "efficiency"}:
-        # poisson ratio n/m is equivalent to binomial n/(n+m)
+    elif uncertainty_type == "poisson-ratio":
         ratio_uncert = np.abs(clopper_pearson_interval(num, num + denom) - ratio)
+    elif uncertainty_type == "efficiency":
+        # poisson ratio n/m is equivalent to binomial n/(n+m)
+        ratio_uncert = np.abs(clopper_pearson_interval(num, denom) - ratio)
     else:
         raise TypeError(
             f"'{uncertainty_type}' is an invalid option for uncertainty_type."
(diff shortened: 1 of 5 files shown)


henryiii (approved):
LGTM, will let @matthewfeickert look it over too just to be sure. :)

src/hist/intervals.py
```diff
@@ -128,7 +128,9 @@ def ratio_uncertainty(
          numerator scaled by the denominator.
          ``"poisson-ratio"`` implements the Clopper-Pearson interval for Poisson
          distributed ``num`` and ``denom``.
```
matthewfeickert, line 130: Should probably mention Nick's https://github.com/scikit-hep/hist/issues/277#issuecomment-892192742 to make clear that these are assumed to be independent. This will compliment the nice explanation added in "efficiency" as well.

matthewfeickert (changes requested):
This is all good and huge thanks @dantrim!  Though we should also update the example in the docs as well. I'm happy to take care of this in the morning once I'm out of meetings though as this fixes things I did wrong previously.

matthewfeickert (approved):
LGTM. Thanks @dantrim!

matthewfeickert:
@all-contributors please add @dantrim for code

matthewfeickert:
Ping @heatherrussell just to give them a heads up on this correction given they were the motivation for PR #266.

PR #281, opened 2021-08-04
Fix Clopper-Pearson and poisson ratio intervals

### LINKED ISSUE #279: [BUG] Fix poisson-ratio implementation for uncertainty intervals
Also, while we're here, since I wrote `# poisson ratio n/m is equivalent to binomial n/(n+m)` in the original coffea code, I think it needs some supporting documents. Strangely I don't recall where I found it, though there is plenty of literature for confidence intervals of poisson ratios, e.g. [here](https://www.newton.ac.uk/files/preprints/ni13040.pdf) or [here](https://ttu-ir.tdl.org/bitstream/handle/2346/59954/31295007034522.pdf?sequence=1), and also this [SO post](https://stats.stackexchange.com/questions/10951/what-is-the-distribution-of-the-ratio-of-two-poisson-random-variables) explaining how to redefine it to even be measurable :) Anyway,

Let n ~ Poisson(lambda_n), m ~ Poisson(lambda_m), then for p = lambda_n / (lambda_n + lambda_m) one can show that the joint probability mass function Poisson(n; lambda_n) * Poisson(m; lambda_m) = Binom(n; n+m, p) * Poisson(n+m; lambda_n + lambda_m). Then the usual [Clopper-Pearson](https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Clopper%E2%80%93Pearson_interval) confidence interval can be used as an estimator of p conditioned on n+m (as is usual). If we instead want to estimate r = lambda_n / lambda_m, then since r = p / (1-p) is monotone increasing, the same confidence interval can be rescaled and still have _at least as good_ coverage properties, though almost surely it begins to overcover.

Though, this means that the `poisson-ratio` implementation is also buggy: it is missing the rescaling:
```python
p_lim = clopper_pearson_interval(num, num + denom)
r_lim = p_lim / (1 - p_lim)  # also need to handle edge cases (corresponds to divide by 0)
ratio_uncert = np.abs(r_lim - ratio) 
```

_Originally posted by @nsmith- in https://github.com/scikit-hep/hist/issues/277#issuecomment-892703615_

nsmith-:
- Fix a bug that erroneously sets the upper Clopper-Pearson interval to 0 when the numerator is zero
- Properly scale the poisson-ratio interval bounds
- Ensure poisson interval lower bound is 0 for 0 observed

Closes #279

### DIFF
--- a/src/hist/intervals.py
+++ b/src/hist/intervals.py
@@ -25,7 +25,7 @@ def __dir__() -> tuple[str, ...]:
 
 
 def poisson_interval(
-    values: np.ndarray, variances: np.ndarray, coverage: float | None = None
+    values: np.ndarray, variances: np.ndarray | None = None, coverage: float | None = None
 ) -> np.ndarray:
     r"""
     The Frequentist coverage interval for Poisson-distributed observations.
@@ -33,14 +33,15 @@ def poisson_interval(
     What is calculated is the "Garwood" interval,
     c.f. https://www.ine.pt/revstat/pdf/rs120203.pdf or
     http://ms.mcmaster.ca/peter/s743/poissonalpha.html.
-    For weighted data, this approximates the observed count by
-    ``values**2/variances``, which effectively scales the unweighted
-    Poisson interval by the average weight.
+    If ``variances`` is supplied, the data is assumed to be weighted, and the
+    unweighted count is approximated by ``values**2/variances``, which effectively
+    scales the unweighted Poisson interval by the average weight.
     This may not be the optimal solution: see https://arxiv.org/abs/1309.1287
     for a proper treatment.
 
-    When a bin is zero, the scale of the nearest nonzero bin is substituted to
-    scale the nominal upper bound.
+    In cases where the value is zero, an upper limit is well-defined only in the case of
+    unweighted data, so if ``variances`` is supplied, the upper limit for a zero value
+    will be set to NaN.
 
     Args:
         values: Sum of weights.
@@ -55,26 +56,20 @@ def poisson_interval(
     # https://github.com/CoffeaTeam/coffea/blob/8c58807e199a7694bf15e3803dbaf706d34bbfa0/LICENSE
     if coverage is None:
         coverage = stats.norm.cdf(1) - stats.norm.cdf(-1)
-    scale = np.ones_like(values)
-    mask = np.isfinite(values) & (values != 0)
-    scale[mask] = variances[mask] / values[mask]
-    if np.sum(values == 0) > 0:
-        missing = np.where(values == 0)
-        available = np.where(mask)
-        if len(available[0]) == 0:
-            raise RuntimeError(
-                "All values are zero! Cannot compute meaningful uncertainties.",
-            )
-        nearest = np.sum(
-            [np.square(np.subtract.outer(d, d0)) for d, d0 in zip(available, missing)]
-        ).argmin(axis=0)
-        argnearest = tuple(dim[nearest] for dim in available)
-        scale[missing] = scale[argnearest]
-    counts = values / scale
-    interval_min = scale * stats.chi2.ppf((1 - coverage) / 2, 2 * counts) / 2.0
-    interval_max = scale * stats.chi2.ppf((1 + coverage) / 2, 2 * (counts + 1)) / 2.0
+    if variances is None:
+        interval_min = stats.chi2.ppf((1 - coverage) / 2, 2 * values) / 2.0
+        interval_min[values == 0.0] = 0.0  # chi2.ppf produces NaN for values=0
+        interval_max = stats.chi2.ppf((1 + coverage) / 2, 2 * (values + 1)) / 2.0
+    else:
+        scale = np.ones_like(values)
+        mask = np.isfinite(values) & (values != 0)
+        np.divide(variances, values, out=scale, where=mask)
+        counts = values / scale
+        interval_min = scale * stats.chi2.ppf((1 - coverage) / 2, 2 * counts) / 2.0
+        interval_min[values == 0.0] = 0.0  # chi2.ppf produces NaN for values=0
+        interval_max = scale * stats.chi2.ppf((1 + coverage) / 2, 2 * (counts + 1)) / 2.0
+        interval_max[values == 0.0] = np.nan
     interval = np.stack((interval_min, interval_max))
-    interval[np.isnan(interval)] = 0.0  # chi2.ppf produces nan for counts=0
     return interval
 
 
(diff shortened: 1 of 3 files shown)


src/hist/intervals.py
```diff
@@ -55,11 +55,12 @@ def poisson_interval(
     # https://github.com/CoffeaTeam/coffea/blob/8c58807e199a7694bf15e3803dbaf706d34bbfa0/LICENSE
     if coverage is None:
         coverage = stats.norm.cdf(1) - stats.norm.cdf(-1)
-    scale = np.empty_like(values)
-    scale[values != 0] = variances[values != 0] / values[values != 0]
+    scale = np.ones_like(values)
+    mask = np.isfinite(values) & (values != 0)
+    scale[mask] = variances[mask] / values[mask]
     if np.sum(values == 0) > 0:
```
nsmith-, line 61: After revisiting this, I'm somewhat dubious if we want to keep the "magic" behavior implemented here. If not for this then the functions would be elementwise (i.e. ufunc-like), which I think makes a bit more sense.

src/hist/intervals.py
```diff
@@ -55,11 +55,12 @@ def poisson_interval(
     # https://github.com/CoffeaTeam/coffea/blob/8c58807e199a7694bf15e3803dbaf706d34bbfa0/LICENSE
     if coverage is None:
         coverage = stats.norm.cdf(1) - stats.norm.cdf(-1)
-    scale = np.empty_like(values)
-    scale[values != 0] = variances[values != 0] / values[values != 0]
+    scale = np.ones_like(values)
+    mask = np.isfinite(values) & (values != 0)
+    scale[mask] = variances[mask] / values[mask]
```
henryiii, line 60:
```suggestion
    np.divide(variances, values, out=scale, where=mask)
```

This will be _much_ faster. Mask selection is slow.

src/hist/intervals.py
```diff
@@ -73,7 +74,7 @@ def poisson_interval(
     interval_min = scale * stats.chi2.ppf((1 - coverage) / 2, 2 * counts) / 2.0
     interval_max = scale * stats.chi2.ppf((1 + coverage) / 2, 2 * (counts + 1)) / 2.0
     interval = np.stack((interval_min, interval_max))
-    interval[interval == np.nan] = 0.0  # chi2.ppf produces nan for counts=0
+    interval[np.isnan(interval)] = 0.0  # chi2.ppf produces nan for counts=0
```
henryiii, line 77: Optional:

```suggestion
    interval = np.nan_to_num(interval, copy=False)  # chi2.ppf produces nan for counts=0
```

This also replaces infinity with very large values, but I think this is safe here.

src/hist/intervals.py
henryiii: I'm generally okay with removing magic. :) @matthewfeickert ?

matthewfeickert: @nsmith- You're saying to just remove this whole logic block and say that if you have values that are zero then you'll need to figure that out? If so, I think that's fine as it is probably better to be explicit with the users about their data/inputs being their responsibility.

src/hist/intervals.py
```diff
@@ -124,26 +125,33 @@ def ratio_uncertainty(
...
-         ``"poisson"`` (default) implements the Poisson interval for the
-         numerator scaled by the denominator.
-         ``"poisson-ratio"`` implements the Clopper-Pearson interval for Poisson
-         distributed ``num`` and ``denom`` where it is assumed that ``num`` and
-         ``denom`` are independent.
-         ``"efficiency"`` implements the Clopper-Pearson interval for Poisson
-         distributed ``num`` and ``denom``, with ``num`` assumed to be a
-         strict subset of ``denom``.
+         ``"poisson"`` (default) implements the Garwood confidence interval for
+         a Poisson-distributed numerator scaled by the denominator.
```
matthewfeickert, line 129: No one is going to know what the Garwood confidence interval is, so if we use this terminology then there should be a direct link in the docs to a citable resource on it.

src/hist/intervals.py
```diff
@@ -124,26 +125,33 @@ def ratio_uncertainty(
...
+         ``"poisson-ratio"`` implements a confidence interval for the ratio assuming
```
matthewfeickert, line 130: > for the ratio

I'd explicitly state the ratio

src/hist/intervals.py
```diff
@@ -124,26 +125,33 @@ def ratio_uncertainty(
...
+         it is an estimator of the ratio of the expected rates from independent Poisson
+         distributions. It does over-cover to a similar degree as the Clopper-Pearson interval
```
matthewfeickert, line 132:
```suggestion
         distributions. It over-covers to a similar degree as the Clopper-Pearson interval
```

src/hist/intervals.py
```diff
@@ -124,26 +125,33 @@ def ratio_uncertainty(
...
+         does for the Binomial efficiency parameter estimate.
+         ``"efficiency"`` implements the Clopper-Pearson confidence interval for the ratio assuming
```
matthewfeickert, line 134: > for the ratio

Would explicitly state the ratio.

tests/test_intervals.py
```diff
@@ -101,46 +101,78 @@ def test_clopper_pearson_interval(hist_fixture):
 
 
 def test_ratio_uncertainty(hist_fixture):
-    hist_1, hist_2 = hist_fixture
+    num, denom = np.meshgrid(np.array([0, 1, 4, 512]), np.array([0, 1, 4, 512]))
```
matthewfeickert, line 104: As you're creating this meshgrid then I don't think you use the `hist_fixture` anywhere else and it should be removed from the test call in the line above.

src/hist/intervals.py
nsmith-: Yeah, specifically this whole indented block starting at L60. I think in the case where a value is zero we should just return a fixed value. Sometimes people propose `1.84` (upper interval assuming unit scale) but I think since we don't have a known scale the best is something like `0` or `nan`. Just another example why frequentist confidence levels are a bit silly.

matthewfeickert (changes requested):
Thanks @nsmith- and sorry about the delay in review (I wanted to give it a look over when I had the time to think about what you were doing and wasn't being kept awake by coffee only).

This overall LGTM along with @henryiii's suggestions and I basically only have documentation comments and then one requested change of removing an unused fixture.

I'll note also that I think the example you gave in Issue #279 is excellent, and I would really enjoy it if in another PR you would add a user guide example to demonstrate the coverage or if you would make an Issue that details this and then review a PR that I would make. :+1:

src/hist/intervals.py
matthewfeickert: SGTM. Also yes. :)

matthewfeickert:
@nsmith- Oh can you also add a short comment in the [`changelog.rst`](https://github.com/scikit-hep/hist/blob/23814cec6adaf60732e04408bb1ee05f9ef08bc8/docs/changelog.rst)?

src/hist/intervals.py
nsmith-: I realize this is preventing nan propagation in the inputs, so I am changing to `interval[values == 0.0] = 0.0` to treat the specific case only.

src/hist/intervals.py
nsmith-: I'll point to the docstring of `poisson_interval` which does have references

src/hist/intervals.py
nsmith-: I made `variances` an optional argument so people can still use `poisson_interval` for unweighted counts and always get an upper limit. Only when variances is supplied do we have this ambiguity.

nsmith-:
Ok, addressed the comments and resolved my outstanding question regarding the "magic".

matthewfeickert (approved):
This all LGTM. :+1: Thanks @nsmith- &mdash; this is great!

I just updated the docstrings for better internal linking (c.f. https://hist--281.org.readthedocs.build/en/281/reference/hist.html#module-hist.intervals) but the content is the same.

henryiii (approved):

matthewfeickert:
@all-contributors please add @nsmith- for code

PR #285, opened 2021-08-10
chore: update Numpy and the matching syntax

### LINKED ISSUE #245: [FEATURE] Update Numpy 1.20 to 1.21 for mypy
We have added types to ndarray, which requires NumPy 1.21 support. We are supposed to make this update for mypy entry in the pre-commit-config.

Probably will hit boost-histogram, too.

amangoel185:
Close #245 when merged.

### DIFF
--- a/src/hist/plot.py
+++ b/src/hist/plot.py
@@ -615,7 +615,7 @@ def _plot_ratiolike(
         if fit_fmt is not None:
             parnames = list(inspect.signature(other).parameters)[1:]
             popt, pcov = bestfit_result
-            perr = np.sqrt(np.diag(pcov))  # type: ignore
+            perr = np.sqrt(np.diagonal(pcov))
 
             fp_label = "Fit"
             for name, value, error in zip(parnames, popt, perr):
--- a/src/hist/plot.py
+++ b/src/hist/plot.py
@@ -125,11 +125,11 @@ def _expr_to_lambda(expr: str) -> Callable[..., Any]:
 
 def _curve_fit_wrapper(
     func: Callable[..., Any],
-    xdata: np.ndarray,
-    ydata: np.ndarray,
-    yerr: np.ndarray,
+    xdata: np.typing.NDArray[Any],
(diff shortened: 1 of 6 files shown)


src/hist/plot.py
```diff
@@ -608,7 +615,7 @@ def _plot_ratiolike(
         if fit_fmt is not None:
             parnames = list(inspect.signature(other).parameters)[1:]
             popt, pcov = bestfit_result
-            perr = np.sqrt(np.diag(pcov))
+            perr = np.sqrt(np.diag(pcov))  # type: ignore
```
henryiii, line 618: Can we switch to using `np.diagonal`? `np.diag` is a Matlab compat function, which is why it's untyped, I think.

src/hist/plot.py
```diff
@@ -651,7 +658,9 @@ def _plot_ratiolike(
             )
 
         elif view == "pull":
-            pulls = (hist_values - compare_values) / hist_values_uncert
+            pulls: np.typing.NDArray[Any] = (
```
henryiii, line 661: We should check and see why this requires a type now - one of the three input values here (hist_values, compare_values, hist_values_uncert) might be missing a type. Don't have to happen in this PR though.

src/hist/plot.py
amangoel185: sure, let me check.

henryiii (approved):

src/hist/plot.py
henryiii: From @amangoel185:

`reveal_type` on the values gives:

```
src/hist/plot.py:661: note: Revealed type is "numpy.ndarray[Any, Any]" # hist_values
src/hist/plot.py:662: note: Revealed type is "numpy.ndarray[Any, numpy.dtype[Any]]" # compare_values
src/hist/plot.py:663: note: Revealed type is "numpy.ndarray[Any, numpy.dtype[Any]]" # hist_values_uncert
```

PR #288, opened 2021-08-11
feat: bulk name setting for AxesTuple

### LINKED ISSUE #274: [FEATURE] bulk name setting for AxesTuple
We should allow the following:

```python
hist.axes.name = ("name1", "name2", ...)
```

Names are read-only so that invalid histograms can't be made, but we can validate in this case, since we know all the names. We should have the same warning if names overlap with fill arguments as when constructing.

LovelyBuggies:
Will close #274 when merged.

### DIFF
--- a/src/hist/axestuple.py
+++ b/src/hist/axestuple.py
@@ -1,9 +1,23 @@
 from __future__ import annotations
 
+import sys
 import warnings
+from collections.abc import Iterable, Iterator
 from typing import Any
 
 from boost_histogram.axis import ArrayTuple, AxesTuple
+
+if sys.version_info < (3, 10):
+    import builtins
+
+    import more_itertools
+
+    def zip(*iterables: Any, strict: bool = False) -> Iterator[tuple[Any, ...]]:
+        if strict:
+            return more_itertools.zip_equal(*iterables)
+        else:
+            return builtins.zip(*iterables)
+
 
 __all__ = ("NamedAxesTuple", "AxesTuple", "ArrayTuple")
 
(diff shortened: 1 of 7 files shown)


LovelyBuggies:
This PR probably gives a feasible solution to allow bulk name setting. But could be in a better manner:

1. To move the `check_name` (duplicability, disabled names) logic outside for both basehist.py and axestuple.py.
2. Should we move this logic to a deeper level (in BH)? - https://github.com/scikit-hep/hist/issues/274#issuecomment-896121377

src/hist/axestuple.py
```diff
@@ -35,6 +36,30 @@ def __getitem__(self, item: Any) -> Any:
 
         return super().__getitem__(item)
 
+    def __setattr__(self, attr: str, values: Any) -> Any:
+        if attr == "name":
```
henryiii, line 40: Can't this just be a normal `@name.setter` ? Is there something I'm missing here? I might try it just to see what happens.

henryiii: I found the reason it can't be a normal setter, and pushed that into boost-histogram. Once 1.1.1 is out, this can be a normal setter. Until then, I've adjusted the PR adapt a similar fix - it's not necessary better or worse, but will be easier to move to boost-histogram 1.1.1+ when that is released.

I've also generalized the values a bit so any arbitrary iterator can be used, and it still checks for length (also applied to boost-histogram in a slightly different way).

henryiii: (will also need to enable Python 3.10 in tests after bh 1.1.1 is out)

Issue #254, opened 2021-07-07
Followup for Stacks


henryiii:
After the initial feature ships post #169, these are some things that we should do (I had several I can't remember at the moment):

* [x] Support showing _histogram_ titles in the legend.
* [ ] Add typing support for titles
* [x] Support Stack.axes as long as we are sure we always will have matching axes (I think we will)
* [x] Add `histoprint` support to `Stack` 

I think `.stack(axis)` will be popular enough that it really doesn't matter if we don't add the `|` shortcut I originally proposed. But we can add that if it seems to be needed.

henryiii:
@andrzejnovak might not agree but I'd be really tempted to retire the `hist_2d.plot()` magic shortcut when one is a category now that `hist_2d.stack(axis).plot()` exists and is much more explicit. Would also simplify the repr.

andrzejnovak:
@henryiii got me, I am convinced the (Cat, Reg) use case is so common it deserves a shorthand even at the expense of explicitness
Worst case scenario there is: user wants a 2D plot, tries `.plot()`, that's not it, tries `.plot2d()`

Side note: I am no longer 100% on the repr needing to project to 1D in this case, maybe 60 now.

LovelyBuggies:
Since we always have matching axes, can we just return the axes of the first histogram in the stack as `s.axes`?

```python
ax = axis.Regular(50, -5, 5, underflow=False, overflow=False, name="X")
h1 = Hist(ax).fill(2 * np.random.normal(size=500) + 2 * np.ones((500,)))
h2 = Hist(ax).fill(2 * np.random.normal(size=500) - 2 * np.ones((500,)))
h3 = Hist(ax).fill(np.random.normal(size=600))
s = Stack(h1, h2, h3)
s.axes # should be (Regular(50, -5, 5, underflow=False, overflow=False, name='X', label='X'))?
```

PR #291, opened 2021-08-11
feat: HistStack followup

LovelyBuggies:
Try to solve #254.


src/hist/stack.py
```diff
@@ -74,7 +81,14 @@ def plot(self, **kwargs: Any) -> list[Hist1DArtists]:
             if all(getattr(h, "name", None) is not None for h in self):
                 kwargs["label"] = [h.name for h in self]  # type: ignore
 
-        return hist.plot.histplot(list(self), **kwargs)  # type: ignore
+        fig = plt.gcf()
+        ax = fig.add_subplot()
+        ret = hist.plot.histplot(list(self), ax=ax, **kwargs)
```
henryiii, line 86: This should be flipped. Someone might pass `ax` already, so we need to support that. What we could/should do is get the axis from `ret`, then set legend entries from each item's name. We also need to make sure the names are set (I think they currently are?).

LovelyBuggies: FYI, I have enabled users to pass axes into `s.plot`, similar to the `_plot_ratiolike` method.

PR #304, opened 2021-09-12
chore: Use `black-jupyter` pre-commit hook over `nbqa-black`

matthewfeickert:
As noted in https://github.com/scikit-hep/hist/pull/276#issuecomment-912956178 there is now a `black[jupyter]` extra that allows for Black to run on Jupyter notebooks. `nbqa-black` does this, but Marco Gorelli contributed this functionality to Black so now the `black-jupyter` hook will be the equivalent hook as running `black` and `nbqa-black`.

c.f. https://github.com/scikit-hep/pyhf/pull/1598 for more details / an example

**Suggested squash and merge commit message**
```
* Use jupyter-black pre-commit hook id from Black
   - black-jupyter will run on Python and pyi files, so it replaces the black hook id as well
* Remove nbqa-black pre-commit hook from nbQA
   - Marco Gorelli essentially contributed nbqa-black to Black in the form of jupyter-black
* Update the pyupgrade version used as an additional_dependencies to match that of the pyupgrade hook
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,8 +38,6 @@ repos:
 - repo: https://github.com/nbQA-dev/nbQA
   rev: 1.1.0
   hooks:
-  - id: nbqa-black
-    additional_dependencies: [black==20.8b1]
   - id: nbqa-pyupgrade
     additional_dependencies: [pyupgrade==2.12.0]
     args: ["--py37-plus"]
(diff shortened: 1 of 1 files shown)


henryiii:
Nice, I knew they added support, but didn't realize the pre-commit integration was so good. I didn't even know types_or was already being used. :) https://github.com/scikit-hep/boost-histogram/pull/638

PR #316, opened 2021-09-20
fix: plot should only require scipy/minuit if required

### LINKED ISSUE #290: [BUG] plot_ratio shouldn't require Minuit?
It seems the try/except is a bit earlier than it needs to be

https://github.com/scikit-hep/hist/blob/6544230fb517853b705c13275df267d6674eb2ac/src/hist/plot.py#L527

@matthewfeickert

henryiii:
Fixes #290.

### DIFF
--- a/src/hist/plot.py
+++ b/src/hist/plot.py
@@ -33,6 +33,8 @@ __all__ = (
     "plot_pie",
 )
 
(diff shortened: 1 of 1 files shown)


matthewfeickert (approved):
LGTM

henryiii:
Thanks!

PR #320, opened 2021-09-21
chore: Python 3.10

henryiii:


henryiii:
Also need to unpin mplhep in tests before release.

PR #338, opened 2021-11-11
chore: minor cleanup to plot.__all__

henryiii:
(pre)Followup to #339.

PR #339, opened 2021-11-12
Fix `ax` arg being ignored by `plot_stack()`

meliache:
The `ax`-argument in `plot_stack` wasn't given to the internally called `histplot` function and then `ax` was overwritten with the axis in the return value of `histplot`. Instead, the argument should be forwarded.

### DIFF
--- a/src/hist/plot.py
+++ b/src/hist/plot.py
@@ -727,7 +727,7 @@ def plot_stack(
         if all(h.name is not None for h in self):
             kwargs["label"] = [h.name for h in self]
 
-    ret = histplot(list(self), **kwargs)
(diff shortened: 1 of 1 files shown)


LovelyBuggies (approved):

henryiii:
This was already done in #338, but I like contributions. :heart: Let's merge this then #338 will be more focused on other unrelated fixes.

meliache:
Thanks for the quick reaction and for merging this PR instead of #338, adding a changelog entry and adding me to the contributors for a duplicate effort :sweat_smile:. Next time I'll take better care to check existing PR's and issues before contributing (and maybe create an issue before doing the PR).

PR #355, opened 2021-11-27
Changing name of hist variable to h

jonas-eschle:
Referencing to #35, implementing with `h` instead of `hist`


jonas-eschle:
Fine from my side

henryiii:
Thanks!

PR #369, opened 2022-01-27
Allow for irregular bin widths in `hist.plot.plot_pull_array`

meliache:
Assuming the axis has an `edges` attribute, I assume it should also have a `widths` attribute, so I thought we can just use that for the pull bar widths instead of calculating our own assuming a regular spacing (which might not be true).

For the patch-width, we just use `right_edge - left_edge`, instead of what was done before, which dividing this difference by the number of pulls and then multiplying them again with the same number.

From git-blame it looks like @matthewfeickert had written the original function so I might mention him.

I should also mention I'm not using this function, I just looked at it because I was writing my own pull plot function and was considering adapting the code for the rectangular patches and then just wondered why `hist.axes[0].widths` isn't used. Thus, it would be good if this could be tested properly.

### DIFF
--- a/src/hist/plot.py
+++ b/src/hist/plot.py
@@ -494,12 +494,12 @@ def plot_pull_array(
     right_edge = __hist.axes.edges[-1][-1]
 
     # Pull: plot the pulls using Matplotlib bar method
-    width = (right_edge - left_edge) / len(pulls)
-    bar_artists = ax.bar(x_values, pulls, width=width, **bar_kwargs)
+    bin_widths = __hist.axes[0].widths
+    bar_artists = ax.bar(x_values, pulls, width=bin_widths, **bar_kwargs)
 
     pp_num = pp_kwargs.pop("num", 5)
     patch_height = max(np.abs(pulls)) / pp_num
-    patch_width = width * len(pulls)
+    patch_width = right_edge - left_edge
     patch_artists = []
     for i in range(pp_num):
         # gradient color patches


meliache:
Now I looked at the file again and `plot_ratio_array` also uses a constant bar width assumption:
https://github.com/scikit-hep/hist/blob/06a485988c3ae3116a5df0b62832af378965bd26/src/hist/plot.py#L428

I didn't use that function and am not sure how it's used and if possibly the ratio doesn't make sense with irregular binnings :shrug: But if you say it's safe I might also change that. I noticed that when I saw the `needs changelog` label and though I should consider which other functions would be affected by this and was thinking whether `plot_ratio_array` would now also allow for non-regular axes.

henryiii:
CC @matthewfeickert, IIRC, on that one?

matthewfeickert:
I'll try to take a look at this before EOD on Monday.

meliache:
Thanks. Before merging I would also update the changelog, I was just waiting for feedback to my [above comment](https://github.com/scikit-hep/hist/pull/369#issuecomment-1023484605), depending on which I might add further changes.

matthewfeickert:
> I'll be trying to get a release out today, FYI.

Once I'm out of my meeting I'll go through this before lunch.

matthewfeickert:
Thanks for the PR @meliache and sorry for letting it get buried in notifications.

> Now I looked at the file again and `plot_ratio_array` also uses a constant bar width assumption:
> 
> https://github.com/scikit-hep/hist/blob/06a485988c3ae3116a5df0b62832af378965bd26/src/hist/plot.py#L428
> 
> I didn't use that function and am not sure how it's used and if possibly the ratio doesn't make sense with irregular binnings :shrug: But if you say it's safe I might also change that. I noticed that when I saw the `needs changelog` label and thought I should consider which other functions would be affected by this and was thinking whether `plot_ratio_array` would now also allow for non-regular axes.

Yeah, I think that's fine. In PR #161 I was indeed making the assumption that the bin widths are constant.

> Even if the calculation doesn't handle irregular bin widths, I still think it's not "wrong" to use the correct edges attributes to compute it (it would just need a check and error in that case).

Nothing should explicitly depend on this, so I would agree that this can be propagated everywhere.

Beyond using the suggested behavior of your PR everywhere, can you also add a test that uses irregular binning?

> I'll be trying to get a release out today, FYI.

@henryiii As I was slow and it is already evening for @meliache I doubt this would be ready to go into a new release today (I have other work that I need to get done before this afternoon as well). I can help @meliache with finishing this tonight or tomorrow if they'd like it, so given the "Hist Serialization and Interactivity" fellow project that you're going to be mentoring will probably have some new releases coming out soon what do you think about getting this PR into a following release, so that I don't slow you down here?

henryiii:
Following release would be fine. Maybe even this one, since https://github.com/scikit-hep/boost-histogram/pull/708 is a serious regression and hist will have to wait till after bh gets fixed. 😞

matthewfeickert:
> Following release would be fine. Maybe even this one, since [scikit-hep/boost-histogram#708](https://github.com/scikit-hep/boost-histogram/pull/708) is a serious regression and hist will have to wait till after bh gets fixed.

Ah very sorry to hear it. :( Hopefully, while serious, it isn't too much of a pain to patch. But sounds good RE: either this release or the next.

@meliache I like your plan, and if you'd like me to help add contributions if you're currently busy I'm happy to do so given that I delayed you here. Of course I'm happy to answer any questions you have and let you finish this one off yourself. :+1:

meliache:
> @henryiii As I was slow and it is already evening for @meliache I doubt this would be ready to go into a new release today (I have other work that I need to get done before this afternoon as well). I can help @meliache with finishing this tonight or tomorrow if they'd like it

Thanks for considering my time zone, evening is good as I don't have any meeting or other work to do, so I can add a couple of things like adapting `plot_ratio_array` and adding a changelog entry and possibly a test  (see below), but after this I'll be a way from my keyboard and probably unresponsive. I never considered any release dates. So, if I'm unresponsive after 21 CET and you want this to be merged before tomorrow, feel free add what you think might be missing.

> Beyond using the suggested behaviour of your PR everywhere, can you also add a test that uses irregular binning?

Good idea, a test would make me more comfortable merging this. The question is what to test. I would assume
1. Test that the changed plotting functions run without exceptions with regular as well as irregular bins
2. Test that we get the expected widths, e.g. how the resulting plots look like. For that I would use a regular axis or a variable axis with a regular binning and check that the bin widths are what we would expect from the constant-width calculation used prior to this PR.  I could either look at the matplotlib artists that the plotting functions return, or I could use something like `unittest.mock` to mock the matplotlib plotting functions and check which arguments they were called with. Tbh the latter seems simpler to me,

I just took a look at your tests directory, probably it should go into `test_plot.py`, either into [`test_general_plot_pull`](https://github.com/scikit-hep/hist/blob/main/tests/test_plot.py#L214) or a separate similar test. The

meliache:
I just realized that `_fit_callable_to_hist` doesn't take into account varying bin widths [when calling `_curve_fit_wrapper`](https://github.com/scikit-hep/hist/blob/main/src/hist/plot.py#L334-L336). Instead of `histogram.values()` they should be somehow normalize by the widths, e.g. like what the `binwnorm` parameter does in the plotting functions. This is one of the things we should fix before claiming that `h.plot_pull` allows for variable axes.

src/hist/plot.py
```diff
@@ -331,8 +331,15 @@ def _fit_callable_to_hist(
 
     # Infer best fit model parameters and covariance matrix
     xdata = histogram.axes[0].centers
+
+    # For axes with varying bin widths correct hist values with widths while maintaing
+    # normalization.
+    bin_widths = histogram.axes[0].width
```
henryiii, line 337:
```suggestion
    bin_widths = histogram.axes[0].widths
```

henryiii:
```console
$ docker run --rm -v$PWD:/hist -w/hist -it quay.io/pypa/manylinux2014_x86_64
# pipx run nox -s regenerate
```

It's easier to compare online, so I'm just pushing the result.

henryiii:
(Sorry for the delay, got nerd-sniped) Looks like this messes with the scaling - should this be normalized by the total width?

src/hist/plot.py
```diff
@@ -331,8 +331,15 @@ def _fit_callable_to_hist(
 
     # Infer best fit model parameters and covariance matrix
     xdata = histogram.axes[0].centers
+
+    # For axes with varying bin widths correct hist values with widths while maintaing
+    # normalization.
+    bin_widths = histogram.axes[0].widths
+    bin_width_fractions = bin_widths / np.sum(bin_widths)
```
henryiii, line 338:
```suggestion
    bin_width_fractions = bin_widths / np.sum(bin_widths) * len(bin_widths)
```

meliache:
> (Sorry for the delay pushing, got nerd-sniped) Looks like this messes with the scaling - should this be normalized by the total width?

Upps, I think I messed this up in 37b6d3df54064f87d5da43a55d8e28f2ebd0b1ce. I didn't test that and somehow assumed that it should be right at that time. Now I think it should work if I replace the `np.sum` in there with a `np.mean`, i.e.
``` python
    h_values_width_corrected = histogram.values() / (bin_widths / np.mean(bin_widths))
 ```
 In that case, in case of a regular binning, the hist values remain unchanged, but if a bin is e.g. twice the mean bin width, the value will be divided by half. Instead of `np.mean(bin_widths)` we could also just do `(right_edge - left_edge) / num_bins`, but I doubt this will be much faster in practice.
 
 Anyway, my local tests now fail again, but I think this is because you changed the baseline. Can I just revert that commit when I think I have fixed it?
 
In case your rushing to review this PR to get it into the new release, I'm not sure I can propagate this all the way up, since this affects more functions that I expected (at least if we want to allow variable axes in general) and I'm still not sure how to best tackle the tests and getting good coverage might need some time for me (though I'd be glad if @matthewfeickert could help with that).

File: src/hist/plot.py [hunk 3]
```diff
@@ -331,8 +331,15 @@ def _fit_callable_to_hist(
 
     # Infer best fit model parameters and covariance matrix
     xdata = histogram.axes[0].centers
+
+    # For axes with varying bin widths correct hist values with widths while
+    # maintaining normalization.
+    bin_widths = histogram.axes[0].widths
+    bin_width_fractions = bin_widths / np.sum(bin_widths) * len(bin_widths)
```
henryiii, line 338:
```suggestion
    bin_width_fractions = bin_widths / np.mean(bin_widths)
```

matthewfeickert:
...eventually making my way back to this after getting some other reviews done. It looks like tests are currently passing (:+1:) but

> and I'm still not sure how to best tackle the tests

I can help add some example tests to cover the cases of irregular binning. I think(?) that `hist` doesn't do Codecov reporting or anything like that (@henryiii if you're game I could add Codecov), but from the diff there's basically nothing that would be uncovered and this is just testing functionality / sniffing for pathological examples / using tests as examples to future us.

henryiii:
Sure, I'd be okay with Codecov. Is that the one that adds a comment to every PR? :/

matthewfeickert:
> Sure, I'd be okay with Codecov. Is that the one that adds a comment to every PR? :/

Yeah. It updates it in place each time there is a coverage change. c.f. https://github.com/scikit-hep/pyhf/pull/1809#issuecomment-1064420561

Not sure if that's desirable or not in your mind.

henryiii:
That's mildly annoying but not terrible. Though I'm guessing there's a way to avoid it: https://github.com/pypa/build/pull/443 for example.

meliache:
To be honest I also got busy with other things, sorry for not replying this week earlier, but I could revisit the tests.

> I can help add some example tests to cover the cases of irregular binning.

Thanks a lot. I think I got overwhelmed by open questions how exactly to do the tests. E.g. should I include asserts in with variable binnings into the existing test functions or create new test functions? What should I best do to avoid code duplication, because necessarily I will basically repeat existing tests with non-regular binnigngs? How fine-grained should the tests be? Should I just test non-crashing behaviour or also generate example images and if yes, I'm not sure what the best procedure is for that. But as always don't let perfection be the enemy of the good

meliache:
I just add two tests:

- `test_image_plot_ratio_variable_axis_with_regular_bins`: Same as `test_image_plot_ratio_callable`, but use an `axis.Variable`with manually set regular bin edges, same as former test with `axis.Regular`, so the plot should look the same as the `axis.Regular` plot, thus I re-use the baseline here.

- `test_image_plot_ratio_callable_bayesian_blocks_bins`: Do a ration plots with variable bin widths. For this I ran the bayesian blocks algorithm on 5000 random numbers because I was too lazy to hand-pick variable bins.
  
Also I added a helper function to create these `plot_ratio_callable` test for different axes/binnings in order to avoid code duplication.

However, I see that for some python versions the plot diffs fail :thinking: . Also I saw that the new baseline for the variable binning looks off, the ratios are larger than the errorbars, so I think I have to adapt the errorbar or ratio calculation, but I decided to upload it anyway to at least have some tests as a starting point. https://github.com/scikit-hep/hist/blob/0ac80f960819da0ed96b7d939601ee687f86506b/tests/baseline/test_image_plot_ratio_callable_bayesian_blocks_bins.png

tests/test_plot.py
```diff
@@ -692,31 +692,92 @@ def test_image_plot_ratio_hist():
...
 
 
+def _make_plot_ratio_callable_test_figure(hist_axis, n_numbers):
+    """
+    Helper function to create a plot_ratio figure with a gaussian model.
+
+    The purpose is to create tests for histograms with different hist_axes, e.g.
+    regular and various variable binnings.
+    """
+    rng = np.random.default_rng(42)
```
meliache, line 702: Here I used a random number generator because `np.random.seed` is officially the recommended way to generate random numbers with a fixed seed, its current documentaition says:

> The best practice is to **not** reseed a BitGenerator, rather to
> recreate a new one. This method is here for legacy reasons.

**However**, I just noticed that the random seed not only affects the numbers to fill the histogram with but also the fit which uses `np.random.normal` [here](https://github.com/meliache/hist/blob/feature/allow-for-irregular-pull-widths/src/hist/plot.py#L347), so this might explain why some tests fail. I think I need to change this back to `np.random.seed(42)`.

meliache:
Some plot tests for older python versions fail and I can't explain this by anything that I did except possible matplotlib changes between versions.  Is there a way I can see the image diff from the CI-tests in github? Or do I have to run local tests for that? nox only tested python3.10 for me because that's the only python version that I have installed on my notebook, I can test other versions by installing older python versions (which I am now doing) or ssh-ing into other computers, but that's a hassle.

meliache:
Now I have arrived where I'm personally happy in this PR, I think I propagated my change up so that irregular bin widths should work for `plot_pull` and `plot_ratio` and I think the tests and the new baseline for `plot_ratio` with irregular bins look good now. I will no stop working on this for now and await further review. If you need more tests please tell me.

Coding-style wise I noticed that I now often repeat the same 2 to 3 lines to normalize the histogram values to the mean bin width. I think this could be factored out into a private helper function but for just 2 lines this doesn't really seam worth it, so I'm not sure. I'll leave it like it's now, except you think it's better to write a function for it.

matthewfeickert:
Thanks. I'll take a look at this tonight after I've finished Snowmass papers. But CI is passing and there are tests now, so my guess is that things are probably :+1:.

src/hist/plot.py
```diff
@@ -672,7 +690,7 @@ def _plot_ratiolike(
 
         elif view == "pull":
             pulls: np.typing.NDArray[Any] = (
-                hist_values - compare_values
+                hist_values_width_corrected - compare_values
             ) / hist_values_uncert
```
meliache, line 694: Btw, `hist_values_uncert` seems unbound if `other` is not a callable or string but a histogram. This is not something I introduced in this PR so if I'm not mistaken this might warrant a separate issue.

The question would be whether we just want to use `np.sqrt(self.variances())` or in case `other` also has variances, if we want to statistically combine both somehow :man_shrugging:

PR #388, opened 2022-03-10
fix: repr limit and fallback on normal repr

### LINKED ISSUE #379: chore: convert to mime bundle for IPython reprs
Let's convert `_repr_svg_` (and any extras) to `_repr_mimebundle_`. See https://github.com/scikit-hep/decaylanguage/pull/213.

### LINKED ISSUE #380: [BUG] Display can crash if histogram really large.
Let's "bail out" of displaying histograms larger than some size (100? 1000?), and just display the same sort of preview that we do for 3D+ histograms. This is much simpler than rebinning arbitrary histograms for display, so at least a good interim measure, and maybe good enough (users can rebin to get it to display, or manually plot).

henryiii:
Closes #379 . Also fixes an issue with an empty histogram breaking by not initializing. This adds a hardcoded limit on the SVG size, 1000 for 1D and 200 in either diminsion for 2D (can increase if we move to using a bitmap). Closes #380. We could make this configurable possibly, but a hard limit is just fine at the moment too - you can always plot, this is just a visual helper.

### DIFF
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -16,7 +16,7 @@ from .axestuple import NamedAxesTuple
 from .axis import AxisProtocol
 from .quick_construct import MetaConstructor
 from .storage import Storage
-from .svgplots import html_hist, svg_hist_1d, svg_hist_1d_c, svg_hist_2d, svg_hist_nd
+from .svgplots import html_hist, svg_hist_1d, svg_hist_1d_c, svg_hist_2d
 from .typing import ArrayLike, Protocol, SupportsIndex
 
 if typing.TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


src/hist/basehist.py
```diff
@@ -87,37 +87,37 @@ def __init__(
             for a in args
         ]
 
-        if args:
```
henryiii, line 90: I just completely removed this if, wasn't needed, and it caused the `super().__init__` to never get called.

henryiii: (If you hide whitespace in the GitHub diff, you'll see basically nothing changed really here except this line)

PR #402, opened 2022-04-06
docs: add documentation from boost-histogram

amangoel185:
In reference to #265


docs/user-guide/comparison.rst
```diff
@@ -0,0 +1,70 @@
+.. _usage-comparison:
+
+Comparison with Boost.Histogram
```
henryiii, line 3: We might want to just link to this page, remove it, or at least prefix it with a comparison to boost-histogram. Users will not be as interested in Boost.Histogram here as they would be in a library called boost-histogram. A page hist vs. boost-histogram is almost what we have in the readme.

docs/user-guide/indexing.rst
```diff
@@ -0,0 +1,31 @@
+.. _usage-indexing:
+
+Indexing
+========
+
+Hist implements the UHI indexing protocol. You can read more about it on the `UHI Indexing <https://uhi.readthedocs.io/en/latest/indexing.html>`_ page.
```
henryiii, line 6: We could also link to the UHI+ page that follows, since Hist follows that one too.

henryiii:
Looks good minus a couple of small comments.

docs/user-guide/comparison.rst
amangoel185: Alright! Should we show both the comparisons on the same page or different ones?

docs/user-guide/indexing.rst
amangoel185: Okay

docs/user-guide/comparison.rst
henryiii: I'd not bother with repeating the comparisons in Hist, users can go to boost-histogram for comparisons with Boost.Histogram. I'd just not have this page at all, I think.

amangoel185: Okay, that makes sense!

henryiii (approved):

PR #414, opened 2022-06-07
fix: allow overwriting labels for plot/overlay

andrzejnovak:


amangoel185 (approved):

src/hist/basehist.py
```diff
@@ -417,6 +417,14 @@ def plot1d(
         cat_ax = self.axes[overlay]
         cats = cat_ax if cat_ax.traits.discrete else np.arange(len(cat_ax.centers))
         d1hists = [self[{overlay: cat}] for cat in cats]
+        if "label" in kwargs:
+            if isinstance(kwargs["label"], list) and len(kwargs["label"]) == len(cats):
```
henryiii, line 421: Does it _have_ to be a list? Couldn't it be any arbitrary collection? Except a string. I assume you are trying not to match on a string?

src/hist/basehist.py
```diff
@@ -417,6 +417,19 @@ def plot1d(
         cat_ax = self.axes[overlay]
         cats = cat_ax if cat_ax.traits.discrete else np.arange(len(cat_ax.centers))
         d1hists = [self[{overlay: cat}] for cat in cats]
+        if "label" in kwargs and hasattr(kwargs["label"], "len"):
```
henryiii, line 420:
```suggestion
        if "label" in kwargs and hasattr(kwargs["label"], "__len__"):
```

henryiii: Actually, this will cause breakage if `"label"` is in kwargs but does not have a length, so better to remove this check and let the final else work.

```suggestion
        if "label" in kwargs:
```

henryiii (approved):

PR #419, opened 2022-06-28
fix: plot1d overlay to work for IntCat axis corrrectly

andrzejnovak:
Reported by @dryrun

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## WIP
+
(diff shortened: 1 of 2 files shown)


henryiii:
@amangoel185 could you add a test & changelog entry for this?

henryiii:
This needs tests, I believe it's not the correct fix and tests showed that. If no one else gets to it I'll eventually do it, but not for the next release.

henryiii:
Implemented a fix that won't break other types of axes.

PR #450, opened 2022-09-25
chore: remove redundant MANIFEST related metadata

Saransh-cpp:
I was adding `Pylint` to `vector` and discovered these redundant lines here 😅


henryiii (approved):
Good catch!

PR #453, opened 2022-10-05
chore: minor cleanups from reverb

henryiii:
A few small things. One message is left, but I think it's a bug in reverb.


src/hist/basehist.py
```diff
@@ -233,8 +233,7 @@ def fill(
 
         data = (data_dict[i] for i in range(len(args), self.ndim))
 
-        total_data = tuple(args) + tuple(data)  # Python 2 can't unpack twice
```
henryiii, line 236: Have no idea why this was here. Hist has never supported Python 2.

Issue #470, opened 2022-12-31
[FEATURE] dask histogram backend


lgray:
**Describe the problem, if any, that your feature request is related to**

Right now the https://github.com/dask-contrib/dask-histogram package exposes an interface that is very nearly 1:1 with boost histogram, which I think is the correct scope for this package. However, many users of columnar analysis tools in HEP prefer hist since its UI is more intuitive and comfortable. It would be best to have a sub-module within hist that exposed a dask-histogram backed interface and suite of tools.

**Describe the feature you'd like**

I would like users to be able to do something like:
```python3
import hist.dask as hist
```
or
```python3
import hist

h = hist.Hist(**axes).as_dask()
```

and have access to the hist interface with (nearly) all of its features but backed by computation in dask.

This is very close to practice since `dask_histogram` already very closely follows (and mostly directly uses) the boost-histogram interface, which means that most of what hist does should be easy to adapt.

This will require a few pieces of interface to be smoothened out in dask_histogram since it's not 1:1 with boost-histogram in some cases. (e.g. what is mentioned in the bottom of the comment here: https://github.com/dask-contrib/dask-histogram/issues/35#issuecomment-1368108690)

**Describe alternatives, if any, you've considered**

The alternative is using the dask histogram interface directly, which while effective, is significantly less widely used and less pleasant than hist. This would require significant readoption and generate attrition risk in our user base.

lgray:
Nominally addressed by #471

PR #471, opened 2023-01-10
feat: hist with a dask-histogram backing

lgray:
A possible solution to #470.

Make it so that we can make hist high-level objects with different bases, add dask_histogram based extension.
This requires: https://github.com/dask-contrib/dask-histogram/pull/38 to smooth out interface inconsistencies.

This PR implements a function to generate hist interfaces based on different flavors of boost-histogram (as dask-histogram is). Probably some discussion needs to happen on how this is implemented in hist, I also need to instrument some tests.

Usage looks like:
image

@henryiii @douglasdavis


lgray:
@henryiii I have no idea how to create the signatures for type factories, I'll figure it out but it'll be faster if you can toss me some pointers. :-)

lgray:
I'll remove the typing extension stuff so tests pass.

lgray:
@henryiii "needs changelog" is very opaque. What exactly is expected?

lgray:
I am also rather confused as to why the pre-commit CI is compiling numpy from source.

henryiii:
Don't worry about the needs-changelog tag, that just tells us that there needs to be a changelog entry. It can be done after the PR is merged.

Pre-commit is probably getting an older numpy on a newer Python. We probably pin something that works with our minimum Python (3.7 IIRC), and pre-commit uses 3.11. Needs fixing.

The dynamic generation of classes is a little bothersome (especially from a static type perspective, a class is a type and dynamically generating one means it's not a static type!). We can fake it, of course, not telling the typing system that this is being done. But it might be worth thinking about if there's any other way to do this before jumping in.

lgray:
@henryiii So it may be the case that the `dask_histogram.boost.Histogram` becomes essentially a mixin with a refactor that @douglasdavis is working on. If that's the case then dynamically changing the base is much preferred and we can still keep the `hist.dask` import pattern. 

Right now the `dask_histogram.boost.Histogram` has a piece of local state it carries around which makes things less simple

lgray:
https://github.com/dask-contrib/dask-histogram/pull/38 has changed such that a subclassing version is possible.
However dask objects use the class member `name` and it clashes with `high.Hist` and I needed to change the name attribute of `hist.Hist` to `hname`.

It largely seems to be an internal detail, if there's a way around it I'd prefer not to change things of course.

lgray:
So we were able to get away with subclassing and it works as intended. A `.compute()` or `dask.compute()` yields a `hist.Hist` with appropriate axis labels and such and we are able to use hist filling semantics on `dask_histogram`s. No alteration of base classes or typing is necessary. 

I'll add some tests to check the functionality and write some docs, then I think this should be ready. Is there anything on top of that you'd want included?

lgray:
@henryiii This should be good to go now, please approve the tests to test. 

I've added tests and docs for `hist.dask`. Let me know if you need anything else before merge.

lgray:
Also @LovelyBuggies if you're still helping to maintain this package ^_^

henryiii:
I think it's safe to make this a patch release, it doesn't really touch anything anyone is using in this form. We can just call it experimental in the release notes for now. :)

PR #474, opened 2023-01-31
feat: add `fill_flattened` support

agoose77:
This PR draft is based upon https://gist.github.com/agoose77/f755d96061a36795dda592276d63f506, which prototyped what `fill_flattened` might look like in Hist. The purpose of this feature is to make histogramming with different-structure arrays more convenient.

This implementation is fairy opinionated; it should be possible, for example to support pass in axis arrays by destructuring obejcts _and_ keyword argument. However, in the pursuit of simplicity, I've made these mutually exclusive; either a single structured object should be used, or the `*args` and `**kwargs` are used as expected to map onto axes.


agoose77:
@henryiii this is a WIP, but I'd appreciate any early feedback!

lgray:
This will need corresponding extensions/changes in dask histogram.

lgray:
Actually looking at how this is implemented - no change to dask-histogram should be needed.
Could you add some tests to check this?

lgray:
Could you add a `hist.dask` test for this as well when you have time?

Thanks!

agoose77:
@lgray Is it reasonable _not_ to add dask tests yet? To properly support dask, this PR should implement support for dask arrays, which it currently does not. That is a future goal, but trying to keep things simple for now.

lgray:
Ah, I see. That is fine. It can come later.

agoose77:
@henryiii I marked this as ready for review, and didn't notice you'd removed that flag. Is it because of the failing tests? I'm hoping they'll pass now.

Or, is it not the policy to include features with partial support for Dask?

henryiii:
I added "draft" since you have wip in the title, is that what you mean?

agoose77:
> I added "draft" since you have wip in the title, is that what you mean?

Ah, that would generally be a good indicator :facepalm:. I will update it now that it's not a wip.

henryiii (approved):

Issue #500, opened 2023-06-07
[FEATURE] Having hist functionalities such as a chisq/ndof and KS test etc.


garvitaa:
**Describe the problem, if any, that your feature request is related to**

I would like to compare the shapes of two hists created using coffea by using KS test and chisq/ndof test.
But the scipy implementation available is not

**Describe the feature you'd like**

It would be helpful to have something similar to:
```
KolmogorovTest(hist1, hist2)
```
with options for using Under/Overflow, comparison of normalization etc, and similarly
```
Chi2Test(hist1, hist2)
```

**Describe alternatives, if any, you've considered**
Tried the `scipy.stat.ks_2samp` and `scipy.stat.chisquare`, but they do not work out of the box for histograms.

iasonkrom:
Hi @henryiii, I think I would like to take a shot at this. I'm thinking of methods of `hist.Hist` objects to do one sample tests of the histogram vs a known distribution and methods to do two sample tests vs other `hist.Hist` objects.

PR #503, opened 2023-06-10
wip: add statistical tests between histograms

iasonkrom:
I'm starting an early PR to add statistical tests to histograms after this feature was requested in https://github.com/scikit-hep/hist/issues/500 and I actually believe it is a good and useful feature.
This is an early WIP PR but I'm starting it now in order to make it publicly visible and ask for feedback on the statistics.

### DIFF
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -547,3 +547,23 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
             raise AssertionError(f"Unsupported storage type {storage}")
 
         return super().sum(flow=flow)
+
+    def chisquare_1samp(self) -> str | NotImplementedError:
+        from hist import stats
+
+        return stats.chisquare_1samp(self)
+
+    def chisquare_2samp(self) -> str | NotImplementedError:
+        from hist import stats
+
+        return stats.chisquare_2samp(self)
+
+    def ks_1samp(self) -> str | NotImplementedError:
+        from hist import stats
+
+        return stats.ks_1samp(self)
+
+    def ks_2samp(self) -> str | NotImplementedError:
+        from hist import stats
+
+        return stats.ks_2samp(self)
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -548,22 +548,32 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
 
         return super().sum(flow=flow)
 
-    def chisquare_1samp(self) -> str | NotImplementedError:
+    def chisquare_1samp(
+        self: hist.BaseHist,
+        distribution: str | Callable[..., Any],
+        args: Any = (),
+        kwds: Any = None,
+    ) -> Any:
         from hist import stats
 
-        return stats.chisquare_1samp(self)
+        return stats.chisquare_1samp(self, distribution, args, kwds)
 
-    def chisquare_2samp(self) -> str | NotImplementedError:
+    def chisquare_2samp(self, other: hist.BaseHist) -> Any:
         from hist import stats
 
-        return stats.chisquare_2samp(self)
+        return stats.chisquare_2samp(self, other)
 
-    def ks_1samp(self) -> str | NotImplementedError:
+    def ks_1samp(
+        self: hist.BaseHist,
+        distribution: str | Callable[..., Any],
+        args: Any = (),
+        kwds: Any = None,
+    ) -> Any:
         from hist import stats
 
-        return stats.ks_1samp(self)
+        return stats.ks_1samp(self, distribution, args, kwds)
 
-    def ks_2samp(self) -> str | NotImplementedError:
+    def ks_2samp(self, other: hist.BaseHist) -> Any:
         from hist import stats
 
-        return stats.ks_2samp(self)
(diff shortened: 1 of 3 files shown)


iasonkrom:
I think so far all 4 implemented tests have a decent core functionality. I'm saying this by running toys and looking at the pvalue distributions because they should be uniform(0, 1) under the null hypothesis.
Next steps are to add the option to consider flow bins in the tests and I also think there is a way to take bin errors into consideration when peforming Kolmogorov tests.

iasonkrom:
@nsmith- I don't think having an option to consider flow bins in ks_1samp tests makes sense right?
Cause like where exactly in the flow bins are you gonna compare the empirical CDF with an actual CDF function. The flow bins technically extend to infinity.

henryiii:
I'm thinking having `flow=True` doesn't make ever make sense, how about we just remove that argument for all of them?

src/hist/basehist.py
```diff
@@ -630,3 +630,43 @@ def sum(self, flow: bool = False) -> float | bh.accumulators.Accumulator:
             raise AssertionError(f"Unsupported storage type {storage}")
 
         return super().sum(flow=flow)
+
+    def chisquare_1samp(
+        self,
+        distribution: str | Callable[..., Any],
+        flow: bool = False,
+        args: Any = (),
+        kwds: Any = None,
```
henryiii, line 639:
```suggestion
        *,
        flow: bool = False,
        *args: Any,
        **kwargs: Any,
```

src/hist/basehist.py
```diff
@@ -630,3 +630,43 @@ def sum(self, flow: bool = False) -> float | bh.accumulators.Accumulator:
...
+    ) -> Any:
+        from hist import stats
+
+        return stats.chisquare_1samp(self, distribution, flow, args, kwds)
```
henryiii, line 643:
```suggestion
        return stats.chisquare_1samp(self, distribution, flow, **kwargs)
```

src/hist/basehist.py
henryiii: Though are these needed? What args might one pass through?

henryiii: I think you could do this with `functools.partial` on the function you pass in to the distribution, right? Would be nice to keep our API surface low.

iasonkrom: Yes these are supposed to be used as function arguments for the CDF which may be a cdf from a scipy distribution, a numba-stats distribution or a user-defined CDF. It did it like that because that's similar to the scipy API (`scipy.stats.ks_1samp`). I didn't put much thought into it.

iasonkrom:
> I'm thinking having `flow=True` doesn't make ever make sense, how about we just remove that argument for all of them?

Hmmm, for chi2 tests it does. You are including the observed number of counts in the flow bins in your tests statistic and comparing vs the expected number of counts (in one sample tests) or with the observed number of counts of another histogram (in two sample tests).
For Kolmogorov tests it's iffy. It doesn't make sense to me at all for one sample, and for two sample it's kinda weird. You can see I did some trick there where I sent the flow bin centers to infinity. It is included in ROOT though for two sample tests betwen `TH1s` but we can remove that.

iasonkrom:
@fabriceMUKARAGE Why did you change it from draft?

fabriceMUKARAGE:
@iasonkrom. oooh, that was a mistake I made when checking the unsuccessful checks. My bad, can we bring it back to the draft?

iasonkrom:
Yup it's back to draft. I was just scared about an accidental merge. Don't worry about failing pytests though. I'm changing things in the statistics and therefore the tests fail if I don't update the values there. But the "real" tests are ran offline where I run toy models to see if my statistics are correct. Proper pytests will be written later since I'm tired of chaging them all the time to pass

fabriceMUKARAGE:
I see it's back to draft mode now. That's true, the proper pytests can be done later, thanks for the insights.

src/hist/stats.py
```diff
@@ -0,0 +1,330 @@
...
+        mode = "asymp"
+        warnings.warn(
+            f"Exact ks_2samp calculation not possible with samples sizes "
+            f"{n1} and {n2}. Switching to 'asymp'.",
+            RuntimeWarning,
+            stacklevel=3,
+        )
+
+    if mode == "exact":
+        success, d, pvalue = _attempt_exact_2kssamp(n1, n2, g, d, alternative)
```
henryiii, line 298: FYI, this really worries me. Using a hidden object is a really bad idea - see https://github.com/copier-org/copier/issues/1225 for the most recent time I've seen someone broken by this! Is it possible to avoid this?

iasonkrom: If you mean the function written by scipy, I was trying to avoid writing too many lines since after defining the test statistic from the histograms, performing the test is the same. I could write a similar function and have it here in hist but it would be mainly done by copying code from scipy

iasonkrom:
@henryiii BTW, I was just busy with other things, I’ll get back to this PR soon. Sorry if I’m being slow.

src/hist/stats.py
iasonkrom: I've just added the code in the `stats.py` file instead of importing from `scipy`. Do you think this is resolved?

PR #536, opened 2023-09-13
docs: Clarify sample term required for Mean storage

matthewfeickert:
* Use the same information as provided by boost-histogram in https://github.com/scikit-hep/boost-histogram/pull/876


henryiii:
Thanks!

PR #563, opened 2024-03-23
fix: hist.dask.new to match top-level package

lgray:
This makes it so that `hist.dask` has the equivalent of `hist.new`, which is was missing.


lgray:
@andrzejnovak FYI

PR #601, opened 2025-02-13
feat: allow wildcards in StrCat list indexing

andrzejnovak:
For a demo hist like 
```
import hist
import numpy as np
import string

h = hist.new.Reg(10, 0, 10).StrCat([], growth=True).Weight()
N = 100000
for cat in ["ABC", "BCD", "CDE", "DEF"]:
    h.fill(
        np.random.normal(np.random.uniform(0, 10), 1, N), cat
    )
h.plot2d()
```

image

This allows slicing like 
```
h[:, "*E*"]
```
image

and 
```
h[:, ["*B*", "?C?"]]
```

image


henryiii:
Thought about it a bit to make sure it shouldn't in fact go to loc. I think leaving it as a shortcut here is fine. Just needs a test or two.

andrzejnovak:
@henryiii added a test. I also had to pass axis id to not look for available only for a specific axis and tweak the typing a bit

henryiii:
Thanks!

PR #654, opened 2026-01-03
Improve error message for invalid weighted fill

Rachit931:
### What changed

Adds an early, clearer `ValueError` when `weight` is provided with a length
that does not match the input data passed to `Hist.fill`.

Previously, this case raised a confusing backend error originating from
`boost-histogram`. This change surfaces a more actionable message at the
`hist` API level.

### Scope

- No behavior change for valid inputs
- Only improves error reporting for already-invalid usage

### DIFF
--- a/src/hist/basehist.py
+++ b/src/hist/basehist.py
@@ -305,7 +305,7 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
 
         data = (data_dict[i] for i in range(len(args), self.ndim))
 
-        # ----START. Clearer error for invalid weighted fill :
+        # Quick check for a common mistake with a weighted fill
         if (
             weight is not None
             and args
@@ -316,7 +316,6 @@ class BaseHist(bh.Histogram, metaclass=MetaConstructor, family=hist):
             raise ValueError(
                 "Weight array must match the length of the input data for a given data"
             )
-        # ----END.
 
         return super().fill(*args, *data, weight=weight, sample=sample, threads=threads)
 


src/hist/basehist.py
```diff
@@ -298,6 +298,19 @@ def fill(
...
+            weight is not None
+            and args
+            and hasattr(weight, "__len__")
+            and hasattr(args[0], "__len__")
+            and len(weight) != len(args[0])
+        ):
+            raise ValueError(
+                "Weight array must match the length of the input data for a given data"
+            )
+        # ----END.
```
henryiii:
```suggestion
```

ikrommyd:
This should work right? You don't have to use kwargs to fill. This PR will check the length of weights with length of `testcat1`  and raise an error
```py
import numpy as np
import hist

# Generate random data
x = np.random.standard_normal(size=(200, 2))
weights = np.random.uniform(0.5, 0.75, size=x.shape[0])

# Create histogram
h = hist.Hist(
    hist.axis.StrCategory([], growth=True, name="strcat"),
    hist.axis.Regular(8, -3.5, 3.5, name="x"),
    hist.axis.Regular(7, -3.3, 3.3, name="y"),
    storage=hist.storage.Weight(),
)

# Fill histogram
xT = x.T
h.fill("testcat1", xT[0], xT[1], weight=weights)
h.fill("testcat2", xT[0], xT[1], weight=weights)
```

ikrommyd:
if you pass in kwargs like `h.fill(strcat="testcat1", x=xT[0], y=xT[1], weight=weights)` sure this PR doesn't catch it because it only checks for args.

Rachit931:
Thanks for catching that. I agree that kwargs path isn’t handled, and reverting makes sense as it is just an error-message improvement. Happy to follow up later with a more compatible fix if useful.

ikrommyd:
I think you can improve the error message by checking the length of array like args and kwargs only but maybe it's better to do this in boost-histogram directly. I don't know

PR #657, opened 2026-01-06
fix: deprecation warning from pyparsing in some jobs

henryiii:
Fixing a warning from pyparsing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,7 @@ dask = [
 plot = [
     "matplotlib >=3.8",
     "mplhep >=0.3.33",
+    "pyparsing>=3.0,<3.3; python_version<'3.10'"
 ]
 fit = [
     "scipy >=1.5.4",


henryiii:
pyparsing 3.3 is causing warnings from matplotlib on Python 3.9, since pyparsing makes releases for older Python versions (3.9 is past official EoL!), but matplotlib drops following SPEC 0. (Have I mentioned I really don't like SPEC 0 / NEP 29 driving a dividing wedge into the Python community?)

So we have a couple of options:

* Arbitrarily force pyparsing<3.3 on Python 3.9. That will fix the warning, but 3.9 users might want or eventually need pyparsing 3.3, limits are very problematic. At least it's in an optional extra.
* Just ignore this warning, since whenever pyparsing removes this hopefully they will follow EoL and drop 3.9 first. If not, it will break stuff. That will ignore it everywhere here, though, not just 3.9.
* Drop 3.9 now; it's only a warning for 3.9 users currently.

PR #685, opened 2026-05-06
feat: add chunked histograms

### LINKED ISSUE #684: [FEATURE] ChunkedHist
From @pfackeldey, we could move ChunkedHist from [https://github.com/pfackeldey/histserv](https://github.com/pfackeldey/histserv/blob/main/src/histserv/chunked_hist.py#L143) here. 

<details><summary>:robot: Plan</summary>
<p>

Here is a plan for integrating `ChunkedHist` into `scikit-hep/hist`.

---

## 1. Core Value Proposition

`ChunkedHist` addresses a real pain point: when you have a histogram with **growable categorical axes** (e.g. `StrCategory([], growth=True)`), every `fill()` that hits a new category triggers a full memory reallocation + copy of the entire dense backing array. For large histograms with many categories, this causes severe memory spikes.

`ChunkedHist` avoids this by splitting categorical axes into independent chunks (plain `np.ndarray`s keyed by category tuple). Dense axes still live inside each chunk. Filling only touches the relevant chunk, and merging (`+`/ `+=`) accumulates chunk-by-chunk.

---

### Maps cleanly to hist

| ChunkedHist concept | How it fits hist |
|---|---|
| Constructor: `*axes, storage=None, name="", label=""` | Identical signature to `BaseHist.__init__` |
| `fill(**kwargs)` with name-based args | Identical public API |
| `to_hist()` → `hist.Hist` | Materialization into dense form |
| `from_hist()` classmethod | Factory for round-tripping |
| `+`/ `+=` merging | Natural complement to `Hist` |
| `__getitem__` with `dict[str, value]` | Same pattern as `BaseHist` dict indexing |

### Needs thoughtful adaptation

| Concept | Issue / Decision needed |
|---|---|
| **Serialization** | Original has custom JSON metadata methods for gRPC. hist already has UHI serialization via `to_uhi()` / `from_uhi()`. We should drop custom JSON and rely on `to_hist()` → UHI, or design a native chunked UHI format. |
| **`Hist` class hierarchy** | `ChunkedHist` **cannot** subclass `BaseHist` (it’s not a single `boost_histogram` array). It should be a standalone class that *wraps* a scratch `Hist` internally and exposes a compatible API where possible. |
| **UHI+ indexing on dense axes** | Original `__getitem__` only selects chunk axes. Locators like `loc`, `at`, `rebin`, `sum` make sense for dense axes—but only when materialized via `to_hist()`. Should `ChunkedHist.__getitem__` support them pre-materialization? Probably not for MVP. |
| **Wildcard matching** | Original does not support it. `BaseHist` supports `*` / `?` on `StrCategory`. We can easily add wildcard matching to chunk-axis selection. |
| **Storage validation** | Original rejects `Mean` / `WeightedMean`. This is because the scratch-histogram reuse trick with structured views is tricky. We should keep the validation, with a plan to relax it later. |
| **Transformed Regular axes** | Original rejects `Regular(..., transform=...)`. We should keep this guard. |

---

## 3. Proposed File Layout

```
src/hist/
  __init__.py          # export ChunkedHist
  chunked.py           # new: ChunkedHist, ChunkAxisSpec, helpers
  basehist.py          # (unchanged)
  ...
tests/
  test_chunked.py      # new ~400-600 line test suite
```

No new subpackage needed—`ChunkedHist` is a single-file addition analogous to `stack.py`.

---

### Constructor
```python
class ChunkedHist:
    def __init__(
        self,
        *axes: AxisProtocol,
        storage: Any | None = None,
        name: str = "",
        label: str = "",
    ) -> None: ...
```

**Requirement**: all axes must have names. Categorical axes (`StrCategory`, `IntCategory`) become chunk axes. All other axes become dense axes.

### Core Methods (adapted from original, cleaned up)

| Method | Behavior |
|---|---|
| `fill(self, *, weight=None, sample=None, threads=None, **kwargs)` | Chunk axes must be scalar. Fills scratch dense hist and saves chunk back. |
| `to_hist(self) -> Hist` | Materializes full dense histogram with all known category keys. |
| `from_hist(cls, source: Hist) -> ChunkedHist` | Factory. |
| `__add__` / `__iadd__` | Chunk-wise merge. Accepts `ChunkedHist` or `Hist`. |
| `__getitem__(self, selection: dict[str, ChunkScalar \| Iterable[ChunkScalar]]) -> ChunkedHist` | Select subset of chunks. Supports wildcard `*` for `StrCategory` chunk axes. |
| `chunk_view(self, selection: dict[str, ChunkScalar]) -> np.ndarray` | Get raw dense view for one exact chunk key. |
| `items()` / `keys()` / `__len__()` / `__contains__()` | Iterate over chunks. |
| `empty_like()` / `reset()` | Convenience. |

### Properties

- `axes` → full axes tuple (with chunk axes in their original positions)
- `dense_axes` → non-categorical axes only
- `chunk_axis_names` → tuple of chunk axis names
- `chunk_axes` → tuple of `ChunkAxisSpec`
- `dense_view_shape` / `dense_view_dtype` → shape/dtype of each chunk

### What we **drop** from the original

- `metadata_dict`, `dense_metadata_dict`, `metadata_json`, `dense_metadata_json`, `from_metadata_json` — gRPC-specific serialization not needed in hist.
- `bytes_repr` utility — not core functionality.

### What we could **add later** (out of scope for initial PR)

- `fill` with array-valued chunk axes (dispatch to multiple chunks in one call).
- `ChunkedHist.view()` returning a dict-like proxy.
- `__array__` protocol.

---

## 5. UHI+ Integration — Options

The original explicitly asks: *"only has a minimal version of UHI+ indexing/slicing (not sure if that can be improved?)"*

Here are the realistic options:

| Approach | Pros | Cons |
|---|---|---|
| **A. Keep dict-only `__getitem__`** (exact chunk-key selection, like original) | Simple, unambiguous, avoids confusion with dense-axis locators | No `loc`, `rebin`, `sum` etc. |
| **B. Add `loc` to chunk-axis selection** | `h[{"cat": hist.loc("a")}]` feels consistent | Overkill for categorical exact keys |
| **C. Add wildcard matching** | `h[{"cat": "*E*"}]` is a clear win, reuse `BaseHist._convert_index_wildcards` | Only meaningful for `StrCategory` |
| **D. Allow dense-axis slicing in `__getitem__`** | `h[{"x": slice(2, 5)}]` would trim every chunk | Complex, overlaps with `to_hist()[...]` |

**My recommendation**: **A + C** for the first PR. Exact dict selection on chunk axes, plus wildcard support for `StrCategory`. Dense-axis slicing should be done *after* `to_hist()`; we can document this clearly.

---

## 6. Testing Plan (`tests/test_chunked.py`)

Sections to cover:

1. **Construction**
   - Valid cases: mixed Regular + StrCategory, IntCategory
   - Error: unnamed axes, zero axes, unsupported storage, unsupported transform

2. **Fill**
   - Scalar chunk axes, array dense axes
   - New category triggers new chunk
   - Missing chunk axis raises
   - Array-valued chunk axis raises
   - `weight=` and `sample=` pass-through

3. **Materialization (`to_hist` / `from_hist`)**
   - Round-trip preserves counts
   - Order of categories in materialized axes matches insertion order
   - Empty ChunkedHist → empty Hist with empty categorical axes

4. **Merging (`+` / `+=`)**
   - Compatible merge sums correctly
   - Incompatible axes / storage / name raises

5. **Selection (`__getitem__`)**
   - Single value per axis
   - Multiple values per axis
   - Wildcard on StrCategory
   - Empty selection returns empty ChunkedHist
   - Non-chunk axis raises

6. **Chunk access**
   - `chunk_view()` exact match
   - Missing chunk raises `KeyError`
   - `keys()`, `items()`, `__len__()`, `__contains__()`

7. **Utilities**
   - `empty_like()`, `reset()`, `repr()`

8. **Integration**
   - Verify `hist.ChunkedHist` is importable from top level

---

## 7. Steps to Implement

1. **Create `src/hist/chunked.py`** with:
   - Helper functions (`_validate_supported_axis`, `_validate_supported_storage`, `_normalize_chunk_scalar`, `normalize_chunk_selection`, etc.)
   - `ChunkAxisSpec` dataclass
   - `ChunkedHist` dataclass with all public API
   - Remove gRPC JSON serialization; add `to_hist`/`from_hist`
   - Add wildcards in `__getitem__` via `_convert_index_wildcards` logic adapted from `basehist.py`

2. **Export in `src/hist/__init__.py`**: add `"ChunkedHist"` to `__all__` and import it.

3. **Add `tests/test_chunked.py`** covering all sections above.

4. **Run quality checks**:
   - `ruff check src/hist/chunked.py`
   - `mypy src/hist/chunked.py`
   - `pytest tests/test_chunked.py`

5. **Update docs**: Docstrings follow hist’s doctest style. Optionally add a short section to `docs/user-guide/`.

---

## 8. Open Questions for You

Before I start implementing, let me know your preference on these:

1. **Top-level or subpackage?** Should `from hist import ChunkedHist` work, or do you prefer `from hist.chunked import ChunkedHist` (or both)?

2. **`__getitem__` on a single exact chunk:** Should `h[{"cat": "a"}]` (where `"cat"` is the *only* chunk axis) return a `ChunkedHist` with one chunk, or a plain `hist.Hist`? The original always returns `ChunkedHist`, but returning `Hist` might be more ergonomic.

3. **`fill()` with array-valued chunk axes:** The original requires scalar chunk-axis values. Would you want to support array chunk-axis values (e.g. `h.fill(x=[0.1, 0.2], cat=["a", "b"])`) dispatching to multiple chunks in one call? This is more complex because you’d need to group by chunk key.

4. **Serialization:** Should `ChunkedHist` support UHI serialization directly (e.g. a JSON dict that lists chunks), or is `to_hist()` → `hist.serialization.to_uhi()` sufficient?

5. **Name requirement:** The original requires all axes to have names. Should we keep this strict (like `NamedHist`), or allow unnamed axes and use positional indexing? My recommendation is to keep names mandatory—chunking is inherently name-based.

Let me know your preferences and I’ll proceed with the implementation.

</p>
</details>

henryiii:
Close #684.

<details><summary>:robot: Suggested followups</summary>
<p>

Here are the natural followups, roughly ordered by impact vs. effort:

---

### 1. **Top-level export** (`from hist import ChunkedHist`)
Currently only `from hist.chunked import ChunkedHist` works. Adding it to `hist/__init__.py` is trivial and would make it discoverable.

---

### 2. **Array-valued chunk axes in `fill()`**
Right now `fill()` requires scalar chunk-axis values:
```python
h.fill(x=[0.1, 0.2], cat="a")   # OK
h.fill(x=[0.1, 0.2], cat=["a", "b"])  # raises
```
Supporting array-valued chunk axes would group by chunk key and dispatch to multiple chunks in one call. This is a common user expectation.

**Tradeoff**: More complex because you need to group the dense-axis data by chunk key and call `dense_hist.fill()` per group.

---

### 3. **Native chunked UHI serialization**
Right now round-tripping through JSON/bytes requires `to_hist()` first, which is expensive for large histograms. A native format that serializes chunk metadata + individual chunk arrays would avoid materialization.

---

### 4. **Custom `__getstate__` / `__setstate__`**
For pickle/dill interop. Without this, pickling a `ChunkedHist` won't work correctly (it has unpicklable internal state like the scratch hist reference).

---

### 5. **`Reporter`-style operations: `*`, `-`, `/`, `**`**
Only `+` / `+=` are implemented. Multiplication, subtraction, division could be useful for e.g. weighted subtraction of backgrounds.

---

### 6. **Relax `Mean` / `WeightedMean` storage restriction**
The scratch-histogram reuse trick is trickier with structured storages, but it's solvable with per-field accumulation.

---

### 7. **Support transformed `Regular` axes**
Currently `Regular(..., transform=...)` is rejected. This is just a validation gate that can be lifted once tested.

---

### 8. **Thread-safe filling**
The current `fill()` reuses a single scratch `Hist` per instance. Parallel filling from multiple threads would race on that scratch buffer. Options:
- A thread-local scratch buffer
- A lock around `fill()`
- Remove the scratch buffer entirely and create a temporary `Hist` per fill (slower but simpler)

---

### 9. **`chunk_view()` on missing keys returns zeros instead of raising**
Currently missing chunks raise `KeyError`. Some workflows might prefer getting a zeroed view for missing chunks (like `dict.get()`).

---

### 10. **Documentation page**
A short user-guide section explaining when to reach for `ChunkedHist` vs. plain `Hist` vs. `dask` hist.

---

### My suggestion for priority order
1. Top-level export (trivial, high visibility)
2. Pickle support (needed for real workflows)
3. Array-valued chunk axes (biggest UX improvement)
4. Native serialization (needed for large-scale I/O)
5. Thread safety (needed for parallel analysis)

What resonates with you?

</p>
</details> 

:robot: Assisted-by: Kimi-K2.6

### DIFF
--- a/src/hist/chunked.py
+++ b/src/hist/chunked.py
@@ -1,0 +1,683 @@
+from __future__ import annotations
+
+import fnmatch
+import itertools
+import typing as tp
+from collections.abc import Iterable, Iterator, Mapping
+from dataclasses import dataclass, field
+from typing import Any
+
+import boost_histogram as bh
+import numpy as np
+
+import hist
+
+__all__ = (
+    "ChunkKey",
+    "ChunkScalar",
+    "ChunkedHist",
+    "normalize_chunk_selection",
+)
+
+
+ChunkScalar = str | int
+ChunkKey = tuple[ChunkScalar, ...]
+
+
+def _validate_supported_axis(axis: tp.Any) -> None:
+    if (
+        isinstance(axis, bh.axis.Regular)
+        and getattr(axis, "transform", None) is not None
+    ):
+        raise ValueError("ChunkedHist does not support transformed Regular axes yet")
+
+
+def _validate_supported_storage(storage: tp.Any) -> None:
+    storage_type = type(storage)
+    if storage_type in {bh.storage.Mean, bh.storage.WeightedMean}:
+        raise ValueError(
+            f"ChunkedHist does not support {storage_type.__name__} storage yet"
+        )
+
+
+def _validate_dense_view(
+    view: tp.Any,
+    *,
+    shape: tuple[int, ...],
+    dtype: np.dtype[tp.Any],
+) -> np.ndarray:
+    array = np.asarray(view)
+    if array.shape != shape:
+        raise ValueError(
+            f"dense view shape mismatch: expected {shape}, got {array.shape}"
+        )
+    if array.dtype != dtype:
+        raise ValueError(
+            f"dense view dtype mismatch: expected {dtype}, got {array.dtype}"
+        )
+    return array
+
+
+def _accumulate_dense_view(target: np.ndarray, source: np.ndarray) -> None:
+    if target.dtype.fields is None:
+        target[...] += source
+        return
+    for field_name in target.dtype.names or ():
+        _accumulate_dense_view(target[field_name], source[field_name])
+
+
+def _zero_dense_view(view: np.ndarray) -> None:
+    if view.dtype.fields is None:
+        view.fill(0)
+        return
+    for field_name in view.dtype.names or ():
+        _zero_dense_view(view[field_name])
+
+
+def _normalize_chunk_scalar(value: tp.Any) -> ChunkScalar:
+    if isinstance(value, np.generic):
+        value = value.item()
+    if not isinstance(value, str | int):
+        raise TypeError(
+            f"chunk axis values must normalize to str or int, got {type(value)=}"
+        )
+    return value
+
+
+def _is_scalar_like(value: tp.Any) -> bool:
+    if isinstance(value, str | bytes):
+        return True
+    if np.isscalar(value):
+        return True
+    return bool(isinstance(value, np.ndarray) and value.ndim == 0)
+
+
+def normalize_chunk_selection(
+    selection: Mapping[str, ChunkScalar | tp.Iterable[ChunkScalar]],
+    *,
+    axis_names: Iterable[str],
+    chunk_axis_names: Iterable[str],
+) -> dict[str, tuple[ChunkScalar, ...]]:
+    known_axes = set(axis_names)
+    known_chunk_axes = set(chunk_axis_names)
+    normalized: dict[str, tuple[ChunkScalar, ...]] = {}
+
+    for axis_name, raw_value in selection.items():
+        if axis_name not in known_axes:
+            raise KeyError(f"unknown axis in slice: {axis_name!r}")
+        if axis_name not in known_chunk_axes:
+            raise ValueError(
+                f"slicing only supports chunk axes; {axis_name!r} is dense"
+            )
+        values = (raw_value,) if isinstance(raw_value, str | int) else tuple(raw_value)
+        if not values:
+            raise ValueError(f"slice for axis {axis_name!r} must be non-empty")
+        normalized[axis_name] = tuple(
+            _normalize_chunk_scalar(value) for value in values
+        )
+
+    return normalized
+
+
+@dataclass(slots=True)
+class ChunkAxisSpec:
+    index: int
+    name: str
+    axis_type: type[tp.Any]
+    label: str
+    metadata: tp.Any
+    growth: bool
+    flow: bool
+    known_keys: list[ChunkScalar]
+
+
+@dataclass(slots=True, init=False)
+class ChunkedHist:
+    chunk_axes: list[ChunkAxisSpec]
+    axes: tuple[tp.Any, ...]
+    storage_type: type[tp.Any]
+    name: str
+    label: str
+    _dense_view_shape: tuple[int, ...] = field(init=False)
+    _dense_view_dtype: np.dtype[tp.Any] = field(init=False)
+    _dense_view_nbytes: int = field(init=False)
+    _chunk_axis_names: tuple[str, ...] = field(init=False)
+    _scratch_dense_hist: hist.Hist[Any] = field(init=False)
+    _chunks: dict[ChunkKey, np.ndarray] = field(default_factory=dict)
+
+    def __init__(
+        self,
+        *axes: tp.Any,
+        storage: tp.Any | None = None,
+        name: str = "",
+        label: str = "",
+    ) -> None:
+        """Initialize a chunked histogram.
+
+        Args:
+            *axes: Histogram axes. Named categorical axes become chunk axes.
+            storage: Boost-histogram storage instance. Defaults to ``Double()``.
+            name: Histogram name.
+            label: Histogram label.
+
+        Example:
+            >>> h = ChunkedHist(
+            ...     hist.axis.Regular(10, 0, 1, name="x"),
+            ...     hist.axis.StrCategory([], growth=True, name="cat"),
+            ... )
+            >>> h.fill(x=[0.2, 0.4], cat="a")
+        """
+        if not axes:
+            raise ValueError("ChunkedHist requires at least one axis")
+        if not all(getattr(axis, "name", None) for axis in axes):
+            raise ValueError("all axes must have names")
+        if storage is None:
+            storage = bh.storage.Double()
+        for axis in axes:
+            _validate_supported_axis(axis)
+        _validate_supported_storage(storage)
+
+        self.axes = tuple(axes)
+        self.storage_type = type(storage)
+        self.name = name
+        self.label = label
+        self._chunks = {}
+
+        self.chunk_axes = []
+        dense_axes: list[tp.Any] = []
+        for index, axis in enumerate(self.axes):
+            if isinstance(axis, bh.axis.IntCategory | bh.axis.StrCategory):
+                self.chunk_axes.append(
+                    ChunkAxisSpec(
+                        index=index,
+                        name=axis.name,
+                        axis_type=type(axis),
+                        label=axis.label,
+                        metadata=axis.metadata,
+                        growth=axis.traits.growth,
+                        flow=axis.traits.overflow or axis.traits.underflow,
+                        known_keys=[_normalize_chunk_scalar(key) for key in axis],
+                    )
+                )
+            else:
+                dense_axes.append(axis)
+
+        self._chunk_axis_names = tuple(spec.name for spec in self.chunk_axes)
+
+        self._scratch_dense_hist = hist.Hist(
+            *dense_axes,
+            storage=self.storage_type(),
+            name=self.name,
+            label=self.label,
+        )
+        dense_view = self._scratch_dense_hist.view(flow=True)
+        self._dense_view_shape = dense_view.shape
+        self._dense_view_dtype = dense_view.dtype
+        self._dense_view_nbytes = dense_view.nbytes
+
+    @classmethod
+    def from_hist(
+        cls,
+        source: hist.Hist[Any],
+    ) -> ChunkedHist:
+        """Build a ``ChunkedHist`` from an existing ``hist.Hist``.
+
+        Args:
+            source: Source histogram to copy into chunked storage.
+
+        Returns:
+            A new ``ChunkedHist`` containing the same bin contents.
+
+        Example:
+            >>> source = hist.Hist(
+            ...     hist.axis.Regular(10, 0, 1, name="x"),
+            ...     hist.axis.StrCategory(["a"], growth=True, name="cat"),
+            ... )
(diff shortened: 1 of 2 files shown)


pfackeldey (commented):
Looks good to me!

There are a few small things, I'm not sure what your preference are here @henryiii, e.g., allowing wildcard matching of categories in getitem, or the fix in _save_chunk_view (which sounds like something that should be fixed...).

pfackeldey (approved):
@henryiii from my side this looks good now (CI fails because of pre-commit though?). I have in theory some more ideas for ChunkedHist, but they can be follow up PRs instead. Maybe it should in the docs as well? Mostly I'd say one would want to use ChunkedHist for 2 cases:
1. to save memory when some categorical axis combinations are never filled (because ChunkedHist is sparse at that level)
2. to avoid expensive copies when a workflow involves growing categorical axes a lot (either in fills or in sum-reductions)

henryiii:
:robot: _AI text below from Fable review_ :robot:

# Review: PR #685 — feat: add chunked histograms

## Overview

Adds `ChunkedHist`, a sparse histogram container that stores one dense NumPy array per categorical-axis key combination instead of materializing the full cartesian product. Categorical axes (`IntCategory`/`StrCategory`) become "chunk" axes; the rest stay dense. Fills go through a reusable scratch `Hist`, and `to_hist()`/`from_hist()` convert to/from plain `Hist`. 684 lines of implementation + 673 lines of tests (53 tests, all passing locally). The design is clean and the docstrings with examples are good.

## Critical issues

**1. Growable dense axes silently lose data, corrupt the scratch buffer, and can crash the interpreter.** `_validate_supported_axis` rejects transformed `Regular` axes but not `growth=True` dense axes. When a fill triggers growth, the scratch hist reallocates its buffer, so the pre-fill `dense_view` array goes stale. I reproduced this:

```python
h = ChunkedHist(hist.axis.Regular(4, 0, 1, name="x", growth=True),
                hist.axis.StrCategory([], growth=True, name="cat"))
h.fill(x=[0.5], cat="a")
h.fill(x=[2.5], cat="b")   # triggers growth — "succeeds" but the entry is lost
# scratch view sum is now 1.0 (should be 0) — permanent contamination
h.fill(x=[0.5], cat="c")   # ValueError: shape mismatch: expected (6,), got (13,)
```

One run of this even died with SIGTRAP (exit 133), suggesting the stale view touches freed memory. Fix: reject `traits.growth` on dense axes in `_validate_supported_axis` (chunked.py:27), same as the other "not yet" gates.

**2. `growth=False` categorical axes silently behave as `growth=True`.** `fill()` and `_remember_chunk_key` (chunked.py:297) accept any new key regardless of the axis's growth trait. A plain `Hist` routes unknown categories to flow; `ChunkedHist` invents a new category:

```python
ch = ChunkedHist(..., axis.StrCategory(["a"], name="cat"))  # growth=False
ch.fill(x=[0.5], cat="z")
list(ch.to_hist().axes[1])  # ['a', 'z'] — plain Hist gives ['a'] + flow count
```

Either raise on unknown keys when `growth` is false, or document the divergence deliberately.

**3. `from_hist` silently drops flow content on chunk axes.** The chunk extraction loop (chunked.py:258) only iterates known category indices, so anything in a categorical overflow bin vanishes — `ChunkedHist.from_hist(h).to_hist().sum(flow=True)` returned 1.0 for a source with sum 2.0. Dense-axis flow is preserved fine. At minimum raise if the source has nonzero categorical-flow content; silent count loss is the worst failure mode for a histogram library.

**4. `ruff check` currently fails on this branch — 40 errors, all in the two new files** (18× TRY003, 11× EM102, 7× EM101, 3× TC003, 1× PYI034). `mypy` and `ruff format` pass. The PYI034 one is real: `__iadd__` is annotated `-> ChunkedHist` but should return `Self`, and it also `return NotImplemented` at chunked.py:635, which contradicts the annotation.

## Smaller issues

- **`items()` doesn't actually copy** (chunked.py:445): chunks are already C-contiguous, so `np.ascontiguousarray` returns the internal array itself. Meanwhile `chunk_view()` intentionally returns the live array (a test mutates through it). Pick one contract per method and document it — right now `items()` *looks* defensive but isn't.
- **Double copy in `__getitem__`** (chunked.py:582): `self._chunks[key].copy(order="C")` is immediately copied again inside `_save_chunk_view`. Drop the outer `.copy()`.
- **`empty_like` hardcodes `ChunkedHist(...)`** (chunked.py:436) — use `type(self)(...)` so subclasses survive `__getitem__`/`__add__`.
- **`__repr__` byte formatting** stops at KB; for a feature whose whole point is large histograms, "10485760.0 KB" is likely. A small ladder through MB/GB would help.
- **`bool` sneaks through `_normalize_chunk_scalar`** (`True` is an `int`), giving keys like `(True,)`. Probably harmless but cheap to normalize via `int()`.

## Test coverage

Coverage of the happy paths and error gates is genuinely thorough. Gaps map exactly to the bugs above: no test fills with a growable dense axis, no test fills an unknown key into a non-growth categorical, no round-trip test with flow content on the chunk axis, and no pickle test (the PR description already flags pickle as a known gap — a `pytest.raises`/xfail placeholder would document the current state).

## Verdict

Solid, well-tested foundation, and the PR description honestly catalogs the deliberate scope cuts. But issues 1–3 are all *silent data loss/corruption* paths, which is the one thing a histogram library can't have — each needs either a fix or a loud validation error before this leaves draft. Issue 4 (ruff) presumably needs to be cleaned up anyway before CI goes green.

PR #715, opened 2026-09-09
feat: support arbitrary mutable mappings for ChunkedHist chunks

henryiii:
CC @pfackeldey.

:robot: _AI text below_ :robot:

Add a `chunks=` keyword to `ChunkedHist()`, `from_hist()`, and `empty_like()` that accepts any `MutableMapping` for chunk storage, such as a disk-backed store. Existing entries in a passed mapping are validated and their keys are adopted, so a persisted store can be reopened. The store is exposed as `.chunks`.

`fill()` and `add_dense_view()` now write the chunk back to the mapping after accumulation, so stores that return copies do not lose updates.


pfackeldey (approved):
LGTM!

PR #721, opened 2026-09-14
feat: add hooks for working with graphed-org/graphed

lgray:
Add hooks to hist that allow it to be used with the delayed execution engine "graphed", it has stabilized enough to where it makes sense to support the code upstream and ease installation.

The graphed hooks ride hist's own CI: `graphed`, `graphed-histogram` and `graphed-executors` (all on PyPI at 0.0.2) are entries in the `test` dependency group behind `python_version >= "3.11"` (graphed requires 3.11+), so `tests/test_graphed.py` runs on the 3.11–3.14 `checks` legs and leaves the 3.10 `checks`/`mypy`/`minimums` jobs and the core-only 3.15 leg untouched; the `>=0.0.2` floors keep "Check minimums" off the 0.0.1 releases. One test (`test_uproot_ttree_fill_end_to_end`) stays skipped upstream: it needs `uproot.graphed`, which arrives with scikit-hep/uproot5#1720, and it skips rather than errors on an installed uproot that predates it.

The pre-commit fixes are committed by hand because pre-commit.ci cannot push to a fork owned by an organization.


lgray:
@henryiii this one is good to go now - please have a look

lgray:
@henryiii ping!

lgray:
@henryiii ping!

lgray:
@henryiii I am once again asking you for a review!
image

henryiii:
Seems fine, little hard to review without more context into graphed. A few questions:

* The names exactly match. Is that best? That is, `Hist` vs. `GraphedHist`, etc. Reprs won't show which one you have, for example.
* Given there already is a graphed_histogram, it seems like the integration could live there? I think the only internal API you are using is `_name_to_index`? (I'm fine with it here, just checking)
* `from .. import hist` then `hist.Hist` might be better than `as HistInMemory`
* Maybe there should be a `graphed` extra?
* FWIW I'll drop 3.10 support in a couple of weeks, probably right after the next release.

lgray:
@henryiii The graphed subdir implementation mirrors exactly that of how it was done for dask-histogram. The reason I chose this was so that the semantics are the same and the namespace tells you what mode you're operating in.

i.e. `hist.Hist` and `hist.graphed.Hist` and `hist.dask.Hist` all get you a histogram you create by invoking `Hist.new` that all take the same arguments. To me, this makes switching between modes easier for code since you can just change the namespace and most of the code remains the name. 

I'm also taking precedent here since hist.dask is implemented in the same way.

I would be fine with altering the reprs to show that these are dask or graphed histograms rather than in-memory histograms.

For where this lives - I think it makes the most sense for all the hist extensions to histograms to live in hist, rather than have a diaspora, especially when they are this thin.

I found calling it `HistInMemory` makes the distinction between the delayed object and the final render histogram a bit simpler, but that's just me.

Do we need a graphed extra when people will be using this typically only when they have graphed (or dask in the dask) case already installed? I think the current complaint on missing import is good enough. It makes people stop to think what they are doing.

lgray:
As for graphed:
- https://graphed.readthedocs.io/
- https://graphed-histogram.readthedocs.io/

It is a task-graph engine similar to dask but with better scaling properties and a HEP-opinionated stance.

henryiii (approved):

